PR #3, opened 2024-11-29
Playwright: Make artifacts upload opt-in

xnyo:
Makes the Playwright artifacts upload opt-in and disabled by default.

Example run with artifacts: https://github.com/grafana/clock-panel/actions/runs/12086765029
Example run without artifacts: https://github.com/grafana/clock-panel/actions/runs/12086660665


tolzhabayev (approved):
👍

PR #2, opened 2024-11-28
adding grafana dependency input pass through

tolzhabayev:
testing it right now in clock panel

PR #4, opened 2024-11-29
Playwright: Support Playwright options in cd workflow

xnyo:
Adds the same inputs as #2, but for the cd workflow


.github/workflows/ci.yml
xnyo, line 1: Just formatting this file

tolzhabayev (approved):

PR #6, opened 2024-12-02
Keeping actions up to date automatically

tolzhabayev:
- moving to pin-by-tag instead. Most of the actions are coming from github anyway.
- adding dependabot action to keep the actions up to date

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+# Set update schedule for GitHub Actions
+
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every week
+      interval: "monthly"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,5 +6,4 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      # Check for updates to GitHub Actions every week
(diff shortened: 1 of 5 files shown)


.github/workflows/cd.yml
```diff
@@ -107,7 +107,7 @@ jobs:
         shell: bash
 
       - name: Checkout specified branch
-        uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
+        uses: actions/checkout@v4.2.2 # v4.2.2
```
xnyo, line 110: Do we have to keep the comments or can we just change them to:

```suggestion
        uses: actions/checkout@v4.2.2
```

tolzhabayev: true 
vs35fh1ruef51

tolzhabayev: done

.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+# Set update schedule for GitHub Actions
```
tolzhabayev, line 1: I am not sure if `directory: /` would also work for https://github.com/grafana/plugin-ci-workflows/blob/814f763609866edfb5df79d7e9a1eed7609bb210/actions/plugins/setup/action.yml#L19

but I guess we just have to try it out and see

xnyo (approved):
LGTM!

PR #8, opened 2024-12-02
Chore: some nitpicks

tolzhabayev:


.github/workflows/cd.yml
```diff
@@ -83,7 +83,7 @@ jobs:
     runs-on: ubuntu-latest
 
     outputs:
-      commit-sha: ${{ steps.commit-sha.outputs.sha }}
+      commit-sha: ${{ steps.checkout-specified-branch.outputs.commit }}
```
tolzhabayev, line 86: I _think_ it's the same?

.github/workflows/cd.yml
```diff
@@ -216,13 +213,16 @@ jobs:
             GCOM_PUBLISH_TOKEN_OPS=plugins/gcom-publish-token:ops
             GCOM_PUBLISH_TOKEN_PROD=plugins/gcom-publish-token:prod
 
-      - name: Determine token
+      - name: Determine which token to use
```
tolzhabayev, line 216: Echos the stage just in case it gets resolved incorrectly

actions/plugins/publish/publish/publish.sh
```diff
@@ -121,7 +121,7 @@ sha=$(git rev-parse HEAD)
 popd > /dev/null
 
 # Publish the plugin
-echo "Publishing to $gcom_api_url..."
+echo "Publishing to $gcom_api_url"
```
tolzhabayev, line 124: Removing the three dots as they add confusion in logs, you might think the whole url is cut out

PR #11, opened 2024-12-19
CI: Lint and typecheck frontend, run tests using test:ci

### LINKED ISSUE #9: Frontend Lint?
We are running linters in backend here:
https://github.com/grafana/plugin-ci-workflows/blob/f3f73859169644920d2132bcecfaef8a1cd4e578/actions/plugins/backend/action.yml#L11-L13

But not running any in the frontend - any reason why not? Are they not unified enough?

xnyo:
Adds some frontend checks to the build and lint action, which were not present in Drone:

- Runs `npm run typecheck` and `npm run lint` when building the plugin.
- Changes the command used to run tests with `npm run test:ci`

Fixes #9

### DIFF
--- a/actions/plugins/frontend/action.yml
+++ b/actions/plugins/frontend/action.yml
@@ -8,9 +8,17 @@ runs:
       shell: bash
       run: ${{ github.action_path }}/pm.sh install
 
+    - name: Lint
+      shell: bash
+      run: ${{ github.action_path }}/pm.sh lint
+
+    - name: Typecheck
+      shell: bash
+      run: ${{ github.action_path }}/pm.sh typecheck
+
     - name: Test
       shell: bash
-      run: ./node_modules/.bin/jest --passWithNoTests --maxWorkers 4
+      run: ${{ github.action_path }}/pm.sh test:ci
 
     - name: Build
       shell: bash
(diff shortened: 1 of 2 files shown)


actions/plugins/frontend/action.yml
```diff
@@ -8,9 +8,17 @@ runs:
       shell: bash
       run: ${{ github.action_path }}/pm.sh install
 
+    - name: Lint
+      shell: bash
+      run: ${{ github.action_path }}/pm.sh lint
+
+    - name: Typecheck
+      shell: bash
+      run: ${{ github.action_path }}/pm.sh typecheck
```
xnyo, line 17: Those two were not present in Drone. Is it safe to always attempt to run them? Will/should they always be defined in `package.json`, or do we have to add a check to ensure that the `lint` and `typecheck` are defined in `package.json` ?

They are scaffolded by create-plugin: https://github.com/grafana/plugin-tools/blob/66052204455d8d139ee005d8586e447ed61a213e/packages/create-plugin/templates/common/_package.json#L4-L17

I want to double-check because this is technically a breaking change from Drone

xnyo: @jackw ptal :slightly_smiling_face:

jackw (approved):
Awesome stuff @xnyo ! 🚀

PR #12, opened 2024-12-19
CI: Add Trufflehog secrets scanning for packaged plugin ZIP files

xnyo:
Adds secrets scanning with trufflehog for the packaged plugin zip files.

Example workflow run (with a fake secret): https://github.com/grafana/clock-panel/actions/runs/12655974757/job/35267517194

Fixes https://github.com/grafana/grafana-community-team/issues/224

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -142,7 +142,7 @@ jobs:
           ref: ${{ inputs.branch }}
 
       - name: Setup
-        uses: grafana/plugin-ci-workflows/actions/plugins/setup@giuseppe/trufflehog
+        uses: grafana/plugin-ci-workflows/actions/plugins/setup@main
         with:
           go-version: ${{ inputs.go-version }}
           node-version: ${{ inputs.node-version }}
@@ -184,17 +184,17 @@ jobs:
         shell: bash
 
       - name: Test and build frontend
-        uses: grafana/plugin-ci-workflows/actions/plugins/frontend@giuseppe/trufflehog
+        uses: grafana/plugin-ci-workflows/actions/plugins/frontend@main
 
       - name: Test and build backend
         if: ${{ steps.check-for-backend.outputs.has-backend == 'true' }}
-        uses: grafana/plugin-ci-workflows/actions/plugins/backend@giuseppe/trufflehog
+        uses: grafana/plugin-ci-workflows/actions/plugins/backend@main
 
       # TODO: security scan?
 
       - name: Package universal ZIP
         id: universal-zip
-        uses: grafana/plugin-ci-workflows/actions/plugins/package@giuseppe/trufflehog
+        uses: grafana/plugin-ci-workflows/actions/plugins/package@main
         with:
           universal: "true"
           dist-folder: dist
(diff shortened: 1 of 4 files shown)


actions/plugins/trufflehog/action.yml
```diff
@@ -0,0 +1,21 @@
...
+      It will be scanned recursively.
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Install Trufflehog
+      shell: bash
+      run: |
+        curl -sSfL https://raw.githubusercontent.com/trufflesecurity/trufflehog/main/scripts/install.sh | sh
```
xnyo, line 17: I decided to use `main` here instead of tagged version. In my opinion, there's no need to pin it to specific version (unlike Mage, golangci-lint, etc) because trufflehog attempts to update to the latest version by default when launched anyways/

.github/workflows/ci.yml
```diff
@@ -205,9 +210,15 @@ jobs:
           output-folder: dist-artifacts
           access-policy-token: ${{ env.SIGN_PLUGIN_ACCESS_POLICY_TOKEN }}
 
+      - name: Trufflehog secrets scanning
+        if: ${{ inputs.run-trufflehog == true }}
+        uses: grafana/plugin-ci-workflows/actions/plugins/trufflehog@main
+        with:
+          folder: dist-artifacts
```
xnyo, line 217: This will scan all the packaged zip files before uploading them. Do we also want to run trufflehog on the source code itself?

tolzhabayev: no, one the packaged zip file should be enough at first.

tolzhabayev (commented):
Awesome! Can we add a way to pass exceptions / exclusions? 

It does get it wrong sometimes or some plugins do want to contain things that it considers sensitive - so having a way to override / ignore some warning would be useful.

actions/plugins/trufflehog/action.yml
KristianGrafana: We should stay away from using latest/main whenever possible, especially in CI. Are you certain that it will automatically updated to latest, and it can't be configured not to? Seems like `--no-update` skips updating.

KristianGrafana:
I second:
> It does get it wrong sometimes or some plugins do want to contain things that it considers sensitive - so having a way to override / ignore some warning would be useful.

But looking at the documentation, I see:
> Is there an easy way to ignore specific secrets?
If the scanned source [supports line numbers](https://github.com/trufflesecurity/trufflehog/blob/d6375ba92172fd830abb4247cca15e3176448c5d/pkg/engine/engine.go#L358-L365), then you can add a trufflehog:ignore comment on the line containing the secret to ignore that secrets.

And this is not really feasible in our case. We could show only verified secrets which would decrease the number of secrets found (false positives on high entropy files for example).

Or, we could build a custom [verification server](https://github.com/trufflesecurity/trufflehog?tab=readme-ov-file#verification-server-example-python) where we control which types of secrets that are supposed to be considered valid.

xnyo:
I have pinned the version and arguments to add exclusions with `--include-detectors` and `--exclude-detectors`, please take a look :slightly_smiling_face:

tolzhabayev (approved):
LGTM, let's start with this and see if we (or plugin devs) run into any issues with false positives.

PR #17, opened 2025-01-08
CD: Pass all inputs to ci.yml

### LINKED ISSUE #16: cd.yml should pass all inputs to ci.yml
cd.yml (deploy action) should pass all inputs accepted by ci.yml when invoking it. This is especially important for the Go, Node and golangci-lint versions, otherwise the build might fail (or it can be done with a different versions compared to ci).

Inputs accepted by ci.yml:

https://github.com/grafana/plugin-ci-workflows/blob/63409a180b3cbf42c735711c23dca3e0853895fb/.github/workflows/ci.yml#L58-L72

Arguments passed from cd.yml to ci.yml:

https://github.com/grafana/plugin-ci-workflows/blob/63409a180b3cbf42c735711c23dca3e0853895fb/.github/workflows/cd.yml#L116-L133

xnyo:
Ensures the build done during cd (deployments) is the same as the one done during ci by passing all inputs from cd.yml to ci.yml.

Fixes #16.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -66,6 +66,18 @@ on:
         required: false
         type: boolean
         default: false
+      go-version:
+        description: Go version to use
+        type: string
+        required: false
+      node-version:
+        description: Node.js version to use
+        type: string
+        required: false
+      golangci-lint-version:
+        description: golangci-lint version to use
+        type: string
+        required: false
 
 concurrency:
   group: cd-${{ github.head_ref || github.run_id }}
@@ -111,7 +123,7 @@ jobs:
         id: checkout-specified-branch
         uses: actions/checkout@v4.2.2
         with:
-          ref: ${{ inputs.branch }}      
+          ref: ${{ inputs.branch }}
 
   ci:
     name: CI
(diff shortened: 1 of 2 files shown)


.github/workflows/cd.yml
```diff
@@ -145,19 +172,22 @@ jobs:
     needs:
       - setup
     with:
-      branch: ${{ github.event.inputs.branch }}
-      run-playwright: ${{ github.event.inputs.run-playwright == true }}
+      branch: ${{ inputs.branch }}
+      go-version: ${{ inputs.go-version }}
+      node-version: ${{ inputs.node-version }}
+      golangci-lint-version: ${{ inputs.golangci-lint-version }}
+      run-playwright: ${{ inputs.run-playwright == true }}
```
tolzhabayev, line 179: is this 

`${{ inputs.run-playwright == true }}`

somehow different from just

`${{ inputs.run-playwright }}` ?

xnyo: I *think* I had to switch it to `${{ inputs.run-playwright == true }}` otherwise it would not be considered a boolean and it caused weirdness somewhere else down the line.

It was already like this before the PR.

.github/workflows/cd.yml
```diff
@@ -145,19 +172,22 @@ jobs:
...
       run-playwright-with-grafana-dependency: ${{ inputs.run-playwright-with-grafana-dependency }}
       run-playwright-with-skip-grafana-dev-image: ${{ inputs.run-playwright-with-skip-grafana-dev-image }}
       run-playwright-with-version-resolver-type: ${{ inputs.run-playwright-with-version-resolver-type }}
-      upload-playwright-artifacts: ${{ github.event.inputs.upload-playwright-artifacts == true }}
-      run-trufflehog: ${{ github.event.inputs.run-trufflehog == true }}
+      upload-playwright-artifacts: ${{ inputs.upload-playwright-artifacts == true }}
```
s4kh, line 183: Is changing `github.event.inputs` to just `inputs`(stripping out the `github.event`) related to 

> The workflow will also receive the inputs in the github.event.inputs context. The information in the inputs context and github.event.inputs context is identical except that the inputs context preserves Boolean values as Booleans instead of converting them to strings. The choice type resolves to a string and is a single selectable option.
?

xnyo: Ah, I didn't know that. I only renamed them because I thought it was too verbose to have `github.event`... and `inputs` was working just fine :sweat_smile: . Great catch! I can refactor to account for that

.github/workflows/cd.yml
xnyo: Changed after Sam's comment, I think we can safely get rid of those if we use `inputs` instead of `github.event.inputs` :)

.github/workflows/cd.yml
xnyo: Done!

s4kh (approved):

tolzhabayev (approved):

PR #40, opened 2025-02-17
Support running CI workflows on external forks

### LINKED ISSUE #31: Support running CI workflows on external forks
The current workflows require access to a service account (in CI, mainly for GCS upload). Access is granted through WIF, but it won't run for forks outside the grafana org (forks from external contributors).

The WIF rules are defined here:

https://github.com/grafana/deployment_tools/blob/59c7aa0bf6d046e0fae7f305617098e92ae963c2/terraform/projects/grafanalabs-workload-identity/github-oidc-service-accounts.tf#L71-L73

It would be ideal to allow CI to run for forks (for automated tests & co for PRs), but without access to the service account.

xnyo:
Fixes https://github.com/grafana/plugin-ci-workflows/issues/31

Allows CI to run on external forks by making those changes:
- Disables GCS upload, otherwise the step fails because forks don't have access to the service account
- Disables plugins signing, the GitHub actions artifacts will contain unsigned plugins

Example PR: https://github.com/grafana/clock-panel/pull/247

Example run: https://github.com/grafana/clock-panel/actions/runs/13375181080?pr=247

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -337,7 +337,7 @@ jobs:
     runs-on: ubuntu-latest
 
     # Do not run upload to GCS for PRs from forks (no access)
-    if: ${{ !github.env.IS_FORK }}
+    if: ${{ github.event.pull_request.head.repo.full_name == github.repository }}
 
     permissions:
       contents: read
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -331,6 +336,9 @@ jobs:
     name: Upload to GCS
     runs-on: ubuntu-latest
 
+    # Do not run upload to GCS for PRs from forks (no access)
+    if: ${{ github.event.pull_request.head.repo.full_name == github.repository }}
```
wbrowne, line 340: Can this not be `if: ${{ !env.IS_FORK }}` here?

wbrowne (approved):

PR #42, opened 2025-02-20
Fix vault secrets step skipped for non-fork PRs

xnyo:
Same as #40 (which was reverted in #41), but includes a bugfix for internal PRs.

For internal PRs, the "Get secrets from vault" step was skipped (treated as if it was a fork PR), causing the package step to fail.

The PR has the same changes as #40, plus a bugfix. The diff for the bugfix is: https://github.com/grafana/plugin-ci-workflows/compare/95103bb803c14175a71132455ad6b61d3bc713e8...51f82be

Example failing run: https://github.com/grafana/metrics-drilldown/actions/runs/13411387311/job/37525765368

Example run with this GHA branch for an internal PR: https://github.com/grafana/plugins-drone-to-gha/actions/runs/13434629597?pr=15

Example run with this GHA branch for a fork PR: https://github.com/grafana/clock-panel/actions/runs/13434699105?pr=249

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -158,6 +158,8 @@ env:
   GCS_ARTIFACTS_BUCKET: "integration-artifacts"
   VAULT_INSTANCE: ops
 
+  IS_FORK: ${{ github.event.pull_request.head.repo.full_name != github.repository }}
+
 jobs:
   test-and-build:
     name: Test and build plugin
@@ -233,7 +235,8 @@ jobs:
           universal: "true"
           dist-folder: dist
           output-folder: dist-artifacts
-          access-policy-token: ${{ env.SIGN_PLUGIN_ACCESS_POLICY_TOKEN }}
+          access-policy-token: ${{ env.SIGN_PLUGIN_ACCESS_POLICY_TOKEN || '' }}
+          allow-unsigned: ${{ env.IS_FORK }}
 
       - name: Package os/arch ZIPs
         id: os-arch-zips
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -185,6 +187,7 @@ jobs:
 
       - name: Get secrets from Vault
         id: get-secrets
+        if: ${{ env.IS_FORK != 'true' }}
```
hugohaggmark, line 190: Maybe we should keep it positive
```suggestion
        if: ${{ env.IS_FORK == 'false' }}
```

.github/workflows/ci.yml
```diff
@@ -232,7 +235,8 @@ jobs:
           universal: "true"
           dist-folder: dist
           output-folder: dist-artifacts
-          access-policy-token: ${{ env.SIGN_PLUGIN_ACCESS_POLICY_TOKEN }}
+          access-policy-token: ${{ env.SIGN_PLUGIN_ACCESS_POLICY_TOKEN || '' }}
+          allow-unsigned: ${{ env.IS_FORK }}
```
hugohaggmark, line 239: Maybe we should be super clear here
```suggestion
          allow-unsigned: ${{ env.IS_FORK == 'true' }}
```

.github/workflows/ci.yml
```diff
@@ -241,7 +245,8 @@ jobs:
           universal: "false"
           dist-folder: dist
           output-folder: dist-artifacts
-          access-policy-token: ${{ env.SIGN_PLUGIN_ACCESS_POLICY_TOKEN }}
+          access-policy-token: ${{ env.SIGN_PLUGIN_ACCESS_POLICY_TOKEN || '' }}
+          allow-unsigned: ${{ env.IS_FORK }}
```
hugohaggmark, line 249: Maybe we should be super clear here
```suggestion
          allow-unsigned: ${{ env.IS_FORK == 'true' }}
```

hugohaggmark (approved):
Great work 🎉  

This is the first time I review at a GitHub job in detail so I'm not sure I understood this correctly but I left some suggestions.

Issue #45, opened 2025-02-25
CD: Upload ZIP files for release tags in a separate GCS folder


xnyo:
Due to how SSS currently works with provisioned plugins, we have to upload a separate artifact for tagged releases.

The GCS URL could be: `{BUCKET}/{PLUGIN_ID}/{VERSION}/tags/{PLUGIN_ID}-{TAG}.zip`

Slack thread with discussion: https://raintank-corp.slack.com/archives/C01C4K8DETW/p1740067230244959

xnyo:
IIRC this was actually done in Drone but left over during the GHA migration:

https://github.com/grafana/grafana-drone-extension/blob/411172f4e3681acb2e78d2ed0d42ba8e1fc093c7/pluginjson/pipeline-deploy.go#L142-L153

PR #52, opened 2025-03-12
CD: Fix path for 'latest' artifacts

xnyo:
Fixes some issues with the "latest" artifacts introduced in #45.

Before, the "latest" artifacts contained the version number in them instead of "latest", this PR fixes it.

Also, it uploads the "any" artifact to the `release/latest`, folder e.g.:

https://storage.googleapis.com/integration-artifacts/grafana-datasourcehttpbackendghateststub-datasource/release/latest/grafana-datasourcehttpbackendghateststub-datasource-latest.zip

Rather than just in `release/latest/any`.

These changes are needed for consistency with the old Drone behaviour.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -396,8 +396,6 @@ jobs:
           cp -r /tmp/dist-artifacts /tmp/dist-artifacts-latest
           # Replace the version with "latest" in the zip file name
           find /tmp/dist-artifacts-latest -type f -exec bash -c 'mv "$1" "${1//${{ fromJSON(needs.ci.outputs.plugin).version }}/latest}"' _ {} \;
-          # TODO: remove, debug
-          ls -1 /tmp/dist-artifacts-latest
         shell: bash
 
       - name: Login to Google Cloud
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -391,44 +391,65 @@ jobs:
...
             if [ "${{ matrix.platform }}" == 'any' ]; then
               echo "gcs_upload_glob=$fn*"
+              echo "gcs_upload_glob_latest=$latest_fn*"
             else
               # strip the extension (.zip) and append the platform instead
               echo "gcs_upload_glob=${fn%.*}.${{ matrix.platform }}*"
+              echo "gcs_upload_glob_latest=${latest_fn%.*}.${{ matrix.platform }}*"
             fi
-          } >> "$GITHUB_ENV"
+          } >> "$GITHUB_OUTPUT"
```
xnyo, line 426: The switch from `$GITHUB_ENV` to `$GITHUB_OUTPUT` is not strictly related to this PR, but it's better because those variables don't have to be shared with all steps

wbrowne (approved):

s4kh (approved):

PR #57, opened 2025-03-18
Custom docker / playwright paths

shelldandy:
- Backwards compatible.
- Adds `playwright-docker-compose-file` and `playwright-config` options to the `ci.yml` workflow
- Adds `playwright-config` and `docker-compose-file` options to the `playwright.yml` workflow
- Allows for provided docker compose path and playwright config path to the runner

This allows a bit more customization for the e2e flows for the metrics drilldown squad as described in this living document:

https://docs.google.com/document/d/1f7_VRmp2PXWLgYnLorQD6tRfeEH74_U588NnKfTrHvs/edit?pli=1&tab=t.0

Unblocks: https://github.com/grafana/metrics-drilldown/pull/171

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -34,7 +34,6 @@ on:
       docker-compose-file:
         required: false
         type: string
-        default: docker-compose.yml
         description: Path to the docker-compose file to use for testing
       playwright-config:
         required: false
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -154,14 +154,15 @@ jobs:
       - name: Wait for Grafana to start
         uses: grafana/plugin-actions/wait-for-grafana@main
 
+      - name: Run Dockerized Playwright tests
+        if: ${{ inputs.dockerized-playwright-command }}
+        id: run-tests-in-container
+        run: ${{ inputs.dockerized-playwright-command }}
+
       - name: Run Playwright tests
+        if: ${{ !inputs.dockerized-playwright-command }}
         id: run-tests
-        run: |
-          if [ -n "${{ inputs.dockerized-playwright-command }}" ]; then
-            ${{ inputs.dockerized-playwright-command }}
-          else
-            npx playwright test --config ${{ inputs.playwright-config }}
-          fi
+        run: npx playwright test --config ${{ inputs.playwright-config }}
 
       - name: Upload artifacts
         uses: actions/upload-artifact@v4.6.1
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -116,10 +116,8 @@ jobs:
       - name: Start Grafana
         # add the -f argument only if "inputs.docker-compose-file" is defined
         run: |
-          dcf="${{ inputs.docker-compose-file }}"
-          dc="docker compose ${dcf:+-f "$dcf"}"
-          dc pull
-          GRAFANA_VERSION=${{ matrix.GRAFANA_IMAGE.VERSION }} GRAFANA_IMAGE=${{ matrix.GRAFANA_IMAGE.NAME }} $dc up -d
+         dcf="${{ inputs.docker-compose-file }}"
+         GRAFANA_VERSION=${{ matrix.GRAFANA_IMAGE.VERSION }} GRAFANA_IMAGE=${{ matrix.GRAFANA_IMAGE.NAME }} docker compose ${dcf:+-f "$dcf"} up -d
 
       - name: Wait for Grafana to start
         uses: grafana/plugin-actions/wait-for-grafana@main
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -141,15 +141,17 @@ jobs:
           cd "$(ls -1)"
           mv ./* "$src/dist/"
 
-      - name: Start Grafana
+      - name: Start Grafana Custom Location
+        if: ${{ inputs.docker-compose-file }}
         run: |
-          if [ -n "${{ inputs.docker-compose-file }}" ]; then
-            docker compose -f ${{ inputs.docker-compose-file }} pull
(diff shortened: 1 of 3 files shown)


.github/workflows/playwright.yml
```diff
@@ -31,6 +31,24 @@ on:
...
+      docker-compose-file:
+        required: false
+        type: string
+        default: docker-compose.yml
+        description: Path to the docker-compose file to use for testing
+      playwright-config:
+        required: false
+        type: string
+        default: playwright.config.ts
+        description: Path to the docker-compose file(s) to use for testing. Can be a single file or comma-separated list of files.
```
xnyo, line 43:
```suggestion
        description: Path to the Playwright config file to use for testing
```

.github/workflows/playwright.yml
```diff
@@ -83,8 +115,13 @@ jobs:
         shell: bash
 
       - name: Install Playwright Browsers
+        if: ${{ !inputs.dockerized-playwright-prepare }}
         run: npx playwright install --with-deps chromium
 
+      - name: Prepare Playwright environment in container
+        if: ${{ inputs.dockerized-playwright-prepare }}
+        run: ${{ inputs.dockerized-playwright-prepare }}
```
xnyo, line 123: I don't think we should allow running inputs directly, this could lead to security issues. What's an example of the commad that this `dockerized-playwright-prepare` input should contain?

.github/workflows/playwright.yml
```diff
@@ -104,17 +141,30 @@ jobs:
...
           docker compose pull
           GRAFANA_VERSION=${{ matrix.GRAFANA_IMAGE.VERSION }} GRAFANA_IMAGE=${{ matrix.GRAFANA_IMAGE.NAME }} docker compose up -d
 
       - name: Wait for Grafana to start
         uses: grafana/plugin-actions/wait-for-grafana@main
 
+      - name: Run Dockerized Playwright tests
+        if: ${{ inputs.dockerized-playwright-command }}
+        id: run-tests-in-container
+        run: ${{ inputs.dockerized-playwright-command }}
```
xnyo, line 162: Same here, we shouldn't allow running the user input directly. What's an example of the value for this `dockerized-playwright-command` variable?

.github/workflows/playwright.yml
```diff
@@ -104,17 +141,30 @@ jobs:
...
           mv ./* "$src/dist/"
 
+      - name: Start Grafana Custom Location
+        if: ${{ inputs.docker-compose-file }}
+        run: |
+          docker compose -f ${{ inputs.docker-compose-file }} pull
+          GRAFANA_VERSION=${{ matrix.GRAFANA_IMAGE.VERSION }} GRAFANA_IMAGE=${{ matrix.GRAFANA_IMAGE.NAME }} docker compose -f ${{ inputs.docker-compose-file }} up -d
+
       - name: Start Grafana
+        if: ${{ !inputs.docker-compose-file }}
```
xnyo, line 151: Nit: I would merge these two steps (`Start Grafana Custom Location` and `Start Grafana`) into one and just conditionally add `-f` or not depending on the value of `inputs.docker-compose-file`, something like this:

```yaml
      - name: Start Grafana
        run: |
          $dcf="${{ inputs.docker-compose-file }}"
          # add the -f argument only if "inputs.docker-compose-file" is defined
          $dc="docker compose ${dcf:+-f "$dcf"}"
          $dc pull
          GRAFANA_VERSION=${{ matrix.GRAFANA_IMAGE.VERSION }} GRAFANA_IMAGE=${{ matrix.GRAFANA_IMAGE.NAME }} $dc up -d
```

.github/workflows/playwright.yml
```diff
@@ -31,6 +31,24 @@ on:
         required: false
         type: boolean
         default: false
+      docker-compose-file:
+        required: false
+        type: string
+        default: docker-compose.yml
```
xnyo, line 37: I believe this should default to an empty string rather than `docker-compose.yml`, also because the default docker-compose file scaffolded by create-plugin is called `docker-compose.yaml` (with an a):

https://github.com/grafana/plugin-tools/blob/main/packages/create-plugin/templates/common/docker-compose.yaml

```suggestion

```

xnyo (changes requested):
Thank you for this! Leaving some comments, and I am a bit confused by the two `dockerized-playwright-prepare` and `dockerized-playwright-command` inputs and what their values should be. Would it be possible to get an example of how running the Playwright tests could look like? (example setup)

Based on that, we should find a way to allow running tests in Docker without passing inputs (commands) directly to the steps, because that is a security risk

.github/workflows/playwright.yml
shelldandy: It's a prepare command just like on bare metal we need to run the `Install Playwright Browsers` before running playwright, so in this case this would do a similar thing but for the dockerized one for example we would pass an npm script like this:

```
"e2e:ci:prepare": "docker build --progress=auto -t metrics-drilldown-plugin-e2e:latest -f e2e/docker/Dockerfile.plugin .",
```

.github/workflows/playwright.yml
shelldandy: An example would also be an npm script to actually run playwright dockerized:

```
"e2e:ci": " docker run --network host -v $(pwd)/e2e:/app/e2e -v $(pwd)/src:/app/src metrics-drilldown-plugin-e2e:latest"
```

.github/workflows/playwright.yml
shelldandy: I do understand the security issues, we're also okay with removing the dockerized playwright and just keeping the other enhancements for now

xnyo (commented):
Great work! :raised_hands: I tested it and found an issue with the docker compose command which needs to be fixed:

xnyo (approved):
LGTM, awesome work! :rocket:

PR #58, opened 2025-03-19
Add support for go dependencies on private repositories

dgiagio:
This PR adds support for Go plugins that have dependencies on private GitHub repositories. It enables the CI workflow to access and download Go modules from private repositories during the build process.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -193,6 +193,16 @@ jobs:
           vault_instance: ${{ env.VAULT_INSTANCE }}
           common_secrets: |
             SIGN_PLUGIN_ACCESS_POLICY_TOKEN=plugins/sign-plugin-access-policy-token:token
+            GITHUB_APP_ID=plugins-platform-bot-app:app-id
+            GITHUB_APP_PRIVATE_KEY=plugins-platform-bot-app:private-key
+
+      - name: Generate GitHub token
+        id: generate-github-token
+        uses: actions/create-github-app-token@v1.11.5
+        with:
+          app-id: ${{ env.GITHUB_APP_ID }}
+          private-key: ${{ env.GITHUB_APP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
 
       - name: Spellcheck
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -193,6 +193,16 @@ jobs:
...
+            GITHUB_APP_ID=plugins-platform-bot-app:app-id
+            GITHUB_APP_PRIVATE_KEY=plugins-platform-bot-app:private-key
+
+      - name: Generate GitHub token
+        id: generate-github-token
+        uses: actions/create-github-app-token@v1.11.5
+        with:
+          app-id: ${{ env.GITHUB_APP_ID }}
+          private-key: ${{ env.GITHUB_APP_PRIVATE_KEY }}
+          owner: ${{ github.repository_owner }}
```
xnyo, line 205: I suppose the provided `${{ secrets.GITHUB_TOKEN }}` (https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication)  isn't enough in this case?

actions/plugins/backend/action.yml
```diff
@@ -1,9 +1,18 @@
...
+  github-token:
+    description: GitHub token for downloading dependencies from private repos, if necessary
+    required: true
+
 runs:
   using: composite
   steps:
+    - name: Config git to use GitHub token
+      run: git config --global url."https://oauth2:${{ inputs.github-token }}@github.com/".insteadOf "https://github.com/"
+      shell: bash
```
xnyo, line 14: Do we also need to update `GOPRIVATE`, or is this enough? I remember reading something about it in the Slack thread :thinking:

xnyo (commented):
Thanks for contributing! :raised_hands:  Great work! Just a few comments

.github/workflows/ci.yml
dgiagio: I also tried using the GitHub token `${{ secrets.GITHUB_TOKEN }}`, but unfortunately, the Go dependency download failed with an error message saying the repository was not found.

xnyo: Makes sense :+1: I'm marking this conversation as resolved then!

xnyo (approved):
LGTM! :raised_hands:

wbrowne (approved):

PR #73, opened 2025-04-23
Publish action: Add support for debug logging

xnyo:
Run `set -x` when debug logging are enabled in GitHub Actions

### DIFF
--- a/actions/plugins/publish/publish/publish.sh
+++ b/actions/plugins/publish/publish/publish.sh
@@ -1,4 +1,6 @@
 #!/bin/bash
+env
+
 set -e
 if [ $ACTIONS_STEP_DEBUG == "true" ]; then
     set -x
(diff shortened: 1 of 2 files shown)


wbrowne (approved):

actions/plugins/publish/publish/publish.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
 set -e
+if [ $RUNNER_DEBUG == "1" ]; then
```
hairyhenderson, line 3: This syntax is causing errors:
```
/opt/actions-runner/_work/_actions/grafana/plugin-ci-workflows/main/actions/plugins/publish/publish/publish.sh: line 3: [: ==: unary operator expected
```

Should probably be:

```suggestion
if [ "$RUNNER_DEBUG" == "1" ]; then
```

xnyo: Great catch! https://github.com/grafana/plugin-ci-workflows/pull/86

PR #99, opened 2025-05-19
CD: Allow git tag creation to fail

### LINKED ISSUE #98: CD: Allow git tag creation to fail
If the tag already exists in git (e.g: manually created, or re-triggering a release), the rest of the workflow for creating a github release will fail.

Allow the step to fail, and keep executing the rest of the workflow.

Related to #54

xnyo:
Fix #98

Example run when the tag already exists: https://github.com/grafana/plugins-drone-to-gha/actions/runs/15107621898/job/42459866382 (see logs for the "Create tag" step)


academo (approved):
<3

PR #100, opened 2025-05-19
feat(version-bump-changelog): add version bump changelog generation workflow

L2D2Grafana:
#### GH Workflow to run use the plugins-platform-bot to bump the npm version and generate a changelog.
- `npm run version`
- Optionally generate a changelog with the npm package `generate-changelog` - I did try https://github.com/github-changelog-generator/github-changelog-generator  but it doesn't list out the individual commits
- Commit the changes
- Need to update this to `@main` grafana/plugin-ci-workflows/actions/plugins/version-bump-changelog@l2d2/version-bump-changelog

#### Tested on a fake plugin
- [test-release-gh-actions](https://github.com/grafana/test-release-gh-actions/actions/runs/15078475308)

Plugins are used to following this flow `npm version`, tag creation, `generate changelog` etc before kicking off the release. This workflow attempts to try and automate the pre steps. Many workflows have tried to do this before
This PR is similar to [build plugin](https://github.com/grafana/plugin-actions/blob/main/build-plugin/action.yml#L145) and [do-release](https://github.com/grafana/plugin-validator/blob/main/.github/workflows/do-release.yml). @xnyo I wonder about adding the draft release in this job like the `build plugin` job. Then plugin developers could do all these steps and then publish to the catalogue. Right now drafting the github release happens at the end of releasing the plugin catalogue in prod and the job fails if you've already created the npm version and tag. It's a common pitfall many people will stumbled into.


actions/plugins/version-bump-changelog/action.yml
```diff
@@ -0,0 +1,115 @@
...
+        git add package.json
+        git add CHANGELOG.md || true  # No-op if changelog not generated
+        git commit -m "Bump version to ${{ steps.bump.outputs.new-version }}"
+        git push origin main
+      env:
+        GITHUB_TOKEN: ${{ steps.generate-github-token.outputs.token }}
+
+    - name: Push latest version
+      shell: bash
+      run: git push origin main
```
xnyo, line 111: Do we need this? We already did `git push origin main` at the end of the previous step

actions/plugins/version-bump-changelog/action.yml
```diff
@@ -0,0 +1,115 @@
...
+        # Prepend the header to the top of the file
+        printf "# Changelog\n\n" | cat - CHANGELOG.md > temp && mv temp CHANGELOG.md
+        # Format headers remove dates and unreleased
+        sed -i -E 's/^(#+ [0-9]+\.[0-9]+\.[0-9]+) \(([Uu]nreleased|[0-9]{4}-[0-9]{2}-[0-9]{2})\)/\1/' CHANGELOG.md
+
+      env:
+        GITHUB_TOKEN: ${{ steps.generate-github-token.outputs.token }}
+
+    - name: Commit changes
+      if: always()  # Ensure it runs even if previous steps are skipped
```
xnyo, line 99: Is this needed? The previous steps are skipped with `if` when not relevant, so they shouldn't fail :thinking:

.github/workflows/version-bump-changelog.yml
xnyo, line 1: This file should actually be in the plugin's repository. If you put it here, it will be a workflow for  the `plugin-ci-workflows` repository itself (which is not what we want).

The reference/example workflows currently live in our EngHub docs (e.g.: https://enghub.grafana-ops.net/docs/default/component/grafana-plugins-platform/plugins-ci-github-actions/010-plugins-ci-github-actions/#workflows_1) source code here: https://github.com/grafana/grafana-community-team/tree/main/docs/plugins-ci-github-actions

xnyo (commented):
Great work! Some comments, but I really like this workflow and it was requested quite a few times already so many thanks for taking care of this! :bow: :100: 

Also, I'm not able to access the testing repository you linked in the PR: https://github.com/grafana/test-release-gh-actions/actions/runs/15078475308

I think it's private, can you add me to it so I can take a look at the output, please?

actions/plugins/version-bump-changelog/action.yml
L2D2Grafana: Removed, good call!

actions/plugins/version-bump-changelog/action.yml
L2D2Grafana: Removed!

.github/workflows/version-bump-changelog.yml
L2D2Grafana: Ah my bad, I can add an example there instead.

actions/plugins/version-bump-changelog/action.yml
```diff
@@ -0,0 +1,110 @@
...
+        echo "new-version=${NEW_VERSION}" >> $GITHUB_OUTPUT
+      env:
+        INPUT_VERSION: ${{ inputs.version }}
+        GITHUB_TOKEN: ${{ steps.generate-github-token.outputs.token }}
+    
+    - name: Generate changelog
+      if: ${{ inputs.generate-changelog == 'true' }}
+      shell: bash
+      run: |
+        npm install -g generate-changelog
```
academo, line 76: nit: instead of installing this globally can you call `npx generate-changelog` in the places you need it.

actions/plugins/version-bump-changelog/action.yml
```diff
@@ -0,0 +1,110 @@
...
+          echo "No previous tag found ${{ steps.previous-tag.outputs.previous-tag }}, generating changelog for initial version only: v${{ steps.bump.outputs.new-version }}"
+          # For initial version, get all commits up to HEAD
+          generate-changelog -t HEAD
+        else
+          echo "Generating changelog from ${{ steps.previous-tag.outputs.previous-tag }} to ${{ steps.bump.outputs.new-version }}"
+          generate-changelog -t ${{ steps.previous-tag.outputs.previous-tag }}...${{ steps.bump.outputs.new-version }}
+        fi
+
+        # Remove existing "Changelog" header anywhere in the file
+        sed -i '/^# Changelog$/d' CHANGELOG.md
```
academo, line 89:
```suggestion
        sed -i '/^#\sChangelog$/Id' CHANGELOG.md
```

might be worth to make it case insensitive just in case and add an explicit whitespace matcher

academo (commented):
Nice work @L2D2Grafana !

I only have one concern I'd like @xnyo to confirm but I think the answer is that it doesn't matter for the plugin release itself


The changelog is generated after the version tag is created. which means the changes to CHANGELOG.md won't be part of that version tag.

A way to fix this is to tell npm to not generate the tag `npm version [version] --no-git-tag-version` and later in the commit phase create the tag `git tag $NEW_VERSION` and `git push --follow-tags`

actions/plugins/version-bump-changelog/action.yml
```diff
@@ -0,0 +1,110 @@
...
+        # Format headers remove dates and unreleased
+        sed -i -E 's/^(#+ [0-9]+\.[0-9]+\.[0-9]+) \(([Uu]nreleased|[0-9]{4}-[0-9]{2}-[0-9]{2})\)/\1/' CHANGELOG.md
+
+      env:
+        GITHUB_TOKEN: ${{ steps.generate-github-token.outputs.token }}
+
+    - name: Commit changes
+      shell: bash
+      run: |
+        git add package.json
```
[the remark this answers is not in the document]
L2D2Grafana: I actually did try to remove this and the package.json version bump didn't get committed, so I'm going to leave it.

L2D2Grafana:
> Nice work @L2D2Grafana !
> 
> I only have one concern I'd like @xnyo to confirm but I think the answer is that it doesn't matter for the plugin release itself
> 
> The changelog is generated after the version tag is created. which means the changes to CHANGELOG.md won't be part of that version tag.
> 
> A way to fix this is to tell npm to not generate the tag `npm version [version] --no-git-tag-version` and later in the commit phase create the tag `git tag $NEW_VERSION` and `git push --follow-tags`

I tried creating the version after the changelog and it seems to have worked out pretty well https://github.com/grafana/test-release-gh-actions/actions/runs/15167700969/job/42649724677

actions/plugins/version-bump-changelog/action.yml
```diff
@@ -0,0 +1,113 @@
...
+        git add CHANGELOG.md || true  # No-op if changelog not generated
+        git commit -m "chore(version): bump version to ${{ steps.bump.outputs.new-version }}"
+        git push origin main
```
academo, line 102:
```suggestion
        git tag -a ${{ steps.bump.outputs.new-version }} -m "Release version ${{ steps.bump.outputs.new-version }}"
        git push origin main --tags
```

you can do the next two steps right here

actions/plugins/version-bump-changelog/action.yml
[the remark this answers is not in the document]
academo: that's quite strange. since `npm version` creates a commit with the changes to package.json

academo (approved):
This looks all correct to me.

It is best if you let @xnyo have a final a look at this before you merge it.

xnyo (approved):
LGTM! Awesome work! :rocket:

PR #103, opened 2025-05-22
feat: simplify playwright docker e2e with profiles

shelldandy:
Simplify the playwright action by not needing a separate `docker-compose` for the playwright tests making use of docker profiles:

https://docs.docker.com/compose/how-tos/profiles/

This requires the root docker-compose to have a `playwright` service with a profile tag of `playwright` to run. That way the service won't start when doing a normal `docker compose up` and can only be called on demand when running the e2e tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,7 +42,11 @@ on:
         required: false
         default: true
       run-playwright-docker:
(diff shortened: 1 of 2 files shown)


xnyo (approved):
LGTM! :raised_hands:

PR #110, opened 2025-06-07
feat: allow custom working directory for plugins

sd2k:
This allows the actions to be used when plugins aren't located at the repo root, e.g. for the `grafana-llm-app` and `grafana-ml-app` plugins. You can see it in use [here](https://github.com/grafana/grafana-llm-app/pull/707).

It doesn't yet propagate this setting down to the playwright actions (e.g. the bit that determines Grafana dependency versions). I have a commit which does, but it needs changes to https://github.com/grafana/plugin-actions/tree/main/e2e-version too.

Note to self: before merging I'll need to switch all the nested workflow refs back to `main`!

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -281,7 +281,7 @@ jobs:
 
   ci:
     name: CI
-    uses: grafana/plugin-ci-workflows/.github/workflows/ci.yml@custom-working-directory # zizmor: ignore[unpinned-uses]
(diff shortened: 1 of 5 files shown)


xnyo (approved):
LGTM, awesome work! :raised_hands: 

Thanks for taking care of this!

**Before merging, remember to switch the branches from `@custom-working-directory` to `@main`**

PR #113, opened 2025-06-10
CD: use provided plugin-directory for changelog-path

sd2k:
The new plugin-directory input is used to specify where the plugin lives,
but I hadn't got as far as the changelog action to notice we needed to
also provide the changelog-path input.

This commit does that.


xnyo (approved):
LGTM! :rocket:

PR #119, opened 2025-06-11
feat(playwright): allow custom plugin directories

sd2k:
Similar to recent PRs, this will allow the Playwright workflow to be
run for plugins that are located outside the root of the repository,
such as the grafana-llm-app.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -142,8 +142,9 @@ jobs:
         # add the -f argument only if "inputs.docker-compose-file" is defined
         run: |
           GRAFANA_VERSION=${{ matrix.GRAFANA_IMAGE.VERSION }} GRAFANA_IMAGE=${{ matrix.GRAFANA_IMAGE.NAME }} docker compose ${DOCKER_COMPOSE_FILE:+-f "$DOCKER_COMPOSE_FILE"} up -d
+        working-directory: ${{ inputs.plugin-directory }}
         env:
-          DOCKER_COMPOSE_FILE: ${{ inputs.plugin-directory }}/${{ inputs.docker-compose-file }}
+          DOCKER_COMPOSE_FILE: ${{ inputs.docker-compose-file }}
 
       - name: Wait for Grafana to start
         uses: grafana/plugin-actions/wait-for-grafana@main # zizmor: ignore[unpinned-uses]
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -15,6 +15,11 @@ on:
         description: Plugin version
         type: string
         required: true
+      plugin-directory:
+        description: Directory of the plugin, if not in the root of the repository.
+        type: string
+        required: false
+        default: .
       # https://github.com/grafana/plugin-actions/blob/main/e2e-version/action.yml
       skip-grafana-dev-image:
         default: false
@@ -75,6 +80,7 @@ jobs:
           skip-grafana-dev-image: ${{ inputs.skip-grafana-dev-image }}
           version-resolver-type: ${{ inputs.version-resolver-type }}
           grafana-dependency: ${{ inputs.grafana-dependency }}
+          plugin-directory: ${{ inputs.plugin-directory }}
 
   playwright-tests:
     needs: resolve-versions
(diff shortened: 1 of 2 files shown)


.github/workflows/playwright.yml
```diff
@@ -75,6 +80,7 @@ jobs:
           skip-grafana-dev-image: ${{ inputs.skip-grafana-dev-image }}
           version-resolver-type: ${{ inputs.version-resolver-type }}
           grafana-dependency: ${{ inputs.grafana-dependency }}
+          plugin-directory: ${{ inputs.plugin-directory }}
```
xnyo, line 83: Opened a PR to support this: https://github.com/grafana/plugin-actions/pull/75

.github/workflows/playwright.yml
```diff
@@ -15,6 +15,11 @@ on:
         description: Plugin version
         type: string
         required: true
+      plugin-directory:
+        description: Directory of the plugin, if not in the root of the repository.
+        type: string
+        required: false
+        default: .
```
xnyo, line 22: This input also be added to `ci.yml` and `cd.yml`, and passed down to the `playwright.yml` workflow

.github/workflows/playwright.yml
```diff
@@ -129,13 +136,14 @@ jobs:
...
           PLUGIN_VERSION: ${{ inputs.version }}
+        working-directory: ${{ inputs.plugin-directory }}
 
       - name: Start Grafana
         # add the -f argument only if "inputs.docker-compose-file" is defined
         run: |
           GRAFANA_VERSION=${{ matrix.GRAFANA_IMAGE.VERSION }} GRAFANA_IMAGE=${{ matrix.GRAFANA_IMAGE.NAME }} docker compose ${DOCKER_COMPOSE_FILE:+-f "$DOCKER_COMPOSE_FILE"} up -d
         env:
-          DOCKER_COMPOSE_FILE: ${{ inputs.docker-compose-file }}
+          DOCKER_COMPOSE_FILE: ${{ inputs.plugin-directory }}/${{ inputs.docker-compose-file }}
```
xnyo, line 146: This causes problems if `inputs.plugin-directory` is empty. The `docker compose` command above has some logic to omit `-f ...` if `DOCKER_COMPOSE_FILE` is empty, but now it's never empty (it's `./` instead).

Can you please cherrypick this to fix it: https://github.com/grafana/plugin-ci-workflows/commit/09b7dc4a5197ed44d52e095c2598fe12bee97fe5

.github/workflows/playwright.yml
xnyo, line 1: `playwright-docker.yml` should probably be updated accordingly as well, but it can be done in a follow-up PR

xnyo (commented):
Thanks for taking care of this :raised_hands: Leaving some initial comments, I still have to test it properly :slightly_smiling_face:

.github/workflows/playwright.yml
sd2k: Good spot! Done, thanks :slightly_smiling_face:

.github/workflows/playwright.yml
sd2k: Done in a58e9b9a36cd0008c75c4b8deda17668e2a1ba87.

xnyo (commented):
I tested this with the grafana-llm-app and for some reason the "Move dist artifact" step fails because the folder already exists in `packages/grafana-llm-app/dist` :thinking: I'm not too sure how this has happened

See here: https://github.com/grafana/grafana-llm-app/actions/runs/15730949254/job/44332133994

Adding a `rm -rf dist` before trying to unzip the dist folder from GitHub artifacts fixes it. Can you please add it like so: https://github.com/grafana/plugin-ci-workflows/commit/855fa1079aef8373a97281fc68b3640b466ca87c#diff-7afcd2d8f7b49bda74843f209eefb7b2da45f7e7803bf2e4bd636699b76aa2d3R129

(ignore the echos/tree/ls, they were for debugging)

This change makes Playwright tests run in the LLM app (however they are failing):

https://github.com/grafana/grafana-llm-app/actions/runs/15731331879/job/44333277742

sd2k:
Thanks @xnyo! Done. I'll take a look at those e2e tests :thinking:

xnyo (approved):
LGTM, awesome work! :tada:

PR #130, opened 2025-06-19
Chore: Use get-vault-secrets without exporting env variables

oshirohugo:
Moves the usage of get-vault-secrets to use step outputs instead of global env variables

Part of https://github.com/grafana/grafana-community-team/issues/427


xnyo (approved):
LGTM!

PR #131, opened 2025-06-20
Prevent re-uploading artifacts with the same tag/platform to GCS

academo:
This PR contains three main changes:

* It prevents GCS artifacts from being overwritten [1]
* It allows ops and dev to ignore gcom re-publishing errors when the deploy target is prod. [2]
* Releasing to ops doesn't longer releases to dev. This introduces higher complexity in the pipeline and it is not always the desired behaviour 

Leaving a longer explanation here for future reference

# [1] prevent GCS artifacts overwrite

Overwriting GCS artifacts creates issues on checksums due to several caching mechanisms in cloud. We generally don't want under almost any circumstance that a plugin artifact in GCS gets overwritten. 

The fix: if the artifact already exists (checked by version-name-platform), it skips the uploading step.

Why not fail if the artifact exists?

Because we use the same artifact in all environments, if we were to do a stepped deploy, first to dev, ops and prod. We'll get an error in ops because the artifact was already there from the dev deployment.

# [2] ignore re-publishing errors in dev and ops when deploying to prod. 

We generally don't want any plugin version to be republished. So we will always fail if a plugin tries to republish in prod.

Why do we allow ops and dev to ignore errors on republish?

We only ignore the errors if we target prod. deploying to prod also triggers a deploy to ops and dev. In the case of a stepped release where we released to dev, then ops, then prod, we will find errors when releasing to prod because dev and ops already exist. 

We will always fail a re-publish in prod. 

Will it fail for errors other than republishing?

yes, we only ignore republishing errors.  

# test runs

Success dev release: https://github.com/grafana/plugins-drone-to-gha/actions/runs/15872094097
error dev re-release of the same version: https://github.com/grafana/plugins-drone-to-gha/actions/runs/15872606544
success prod release: https://github.com/grafana/plugins-drone-to-gha/actions/runs/15872264335

in the prod release test, we got a stub error in ops, this is unrelated to the pr changes. 

It proves both that we only ignore republishing errors and that we correctly ignore republishing errors (in the case of dev)

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -497,7 +497,7 @@ jobs:
           gcloud-auth-token: ${{ steps.gcloud.outputs.id_token }}
 
       - name: Publish to catalog
-        uses: grafana/plugin-ci-workflows/actions/plugins/publish/publish@academo/prevent-republish # zizmor: ignore[unpinned-uses]
+        uses: grafana/plugin-ci-workflows/actions/plugins/publish/publish@main # zizmor: ignore[unpinned-uses]
         with:
           zips: ${{ needs.ci.outputs.gcs-zip-urls-commit }}
           environment: ${{ matrix.environment }}
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -236,10 +236,10 @@ jobs:
     steps:
       - name: Check environment
         run: |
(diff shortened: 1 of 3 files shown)


beejeebus (approved):

tolzhabayev (commented):
This might be problematic if the workflows are run for single environments before going for prod (which is what we actually want people to do).

So if someone deploys first to dev and then ops and then runs prod.
Prod will then run it again for dev and ops and with this change it might fail which is not what we want.

beejeebus (approved):
LGTM.

xnyo (approved):
LGTM, awesome work! :tada:  Thank you again so much for taking care of this!

PR #135, opened 2025-06-30
CD: Allow publishing to more than one environment via inputs

### LINKED ISSUE #134: CD: Allow deploying to multiple environments at the same time
#131 changed the behaviour of deployment to be single-stage (except when publishing to prod).

This means publishing to "ops" will target "ops" only (it used to be dev + ops).

Some provisioned plugins (e.g.: logs-drilldown) want to run CD in dev + ops. This is not possible anymore after merging #131. We should keep the default behavior as-is, but allow publishing to dev + ops if the plugin specifies so.

E.g.:

```
environment: dev,ops
```

xnyo:
Allow to deploy to multiple environments by passing a comma-separated string to `environments` (catalog + Argo).

The default behaviour is still deploying to one env when targeting dev and ops, and deploy to all envs when targeting prod.

However you can now pass: `dev,ops` to publish to `dev + ops` (effectively, the way `ops` behaved before #131)

This is handy for provisioned plugins that want to continuously deploy main to ops, like the Logs Drilldown app.

Fixes #134

**Test runs**:

Dev only: https://github.com/grafana/plugins-drone-to-gha/actions/runs/15995008964

Ops only: https://github.com/grafana/plugins-drone-to-gha/actions/runs/15995177803

Dev + Ops: https://github.com/grafana/plugins-drone-to-gha/actions/runs/15995366988

Prod: https://github.com/grafana/plugins-drone-to-gha/actions/runs/15995609109

(Ignore the failures, these are from GCOM, just see the results of the "environments" matrix)

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -365,9 +365,9 @@ jobs:
         with:
           script: |
             // Helper to get input/env values
-            const getInput = (name, fallback = '') => {
-              return process.env[name] !== undefined ? process.env[name] : fallback;
-            };
+            function getInput(name) {
+              return process.env[name] !== undefined ? process.env[name] : '';
+            }
 
             const ENVIRONMENT = getInput('ENVIRONMENT');
             const DOCS_ONLY = getInput('DOCS_ONLY');
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -365,11 +365,11 @@ jobs:
         with:
           script: |
             // Helper to get input/env values
-            function getInput(name) {
-              return process.env[name] !== undefined ? process.env[name] : '';
+            function getInput(name, fallback = '') {
+              return process.env[name] !== undefined ? process.env[name] : fallback;
             }
 
-            const ENVIRONMENT = getInput('ENVIRONMENT');
+            const ENVIRONMENT = getInput('ENVIRONMENT', 'none');
             const DOCS_ONLY = getInput('DOCS_ONLY');
             const HAS_BACKEND = getInput('HAS_BACKEND');
 
@@ -388,7 +388,7 @@ jobs:
             }
 
             // Special case to skip deployment
-            if (!ENVIRONMENT || ENVIRONMENT === 'none') {
+            if (ENVIRONMENT === 'none') {
               core.setOutput('environments', JSON.stringify([]));
               core.setOutput('publish-docs', 'false');
               return;
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -397,7 +397,7 @@ jobs:
               echo 'environments=[]'
               echo 'publish-docs=true'
             } >> "$GITHUB_OUTPUT"
-          elif [ "${environments}" == 'prod' ]; then
+          elif echo "${environments}" | jq 'index("prod") != null'; then
             # Special case: 'prod' deploys to all environments, to keep the catalogs in sync.
             environments="$allowed_environments"
           elif [ "${environments}" == "[]" ]; then
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -241,10 +241,10 @@ jobs:
     steps:
       - name: Check environment
         run: |
-          if [ -n "${ENVIRONMENT}" ] && [ "${ENVIRONMENT}" != 'none' ] && [ "${ENVIRONMENT}" != 'dev' ] && [ "${BRANCH}" != 'main' ]; then
-            echo "Only 'dev' environment is allowed for non-main branches."
-            exit 1
-          fi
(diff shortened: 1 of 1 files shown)


.github/workflows/cd.yml
```diff
@@ -236,10 +241,10 @@ jobs:
     steps:
       - name: Check environment
         run: |
-          if [ -n "${ENVIRONMENT}" ] && [ "${ENVIRONMENT}" != 'none' ] && [ "${ENVIRONMENT}" != 'dev' ] && [ "${BRANCH}" != 'main' ]; then
-            echo "Only 'dev' environment is allowed for non-main branches."
-            exit 1
-          fi
+          #if [ -n "${ENVIRONMENT}" ] && [ "${ENVIRONMENT}" != 'none' ] && [ "${ENVIRONMENT}" != 'dev' ] && [ "${BRANCH}" != 'main' ]; then
```
academo, line 244: remove commented code.

xnyo: Yes you are correct, I only added this for testing and forgot to remove it

.github/workflows/cd.yml
```diff
@@ -374,35 +379,36 @@ jobs:
             exit 0
           fi
 
+          allowed_environments='["dev", "ops", "prod"]'
```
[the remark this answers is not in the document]
xnyo: Tested all the scenarios again with the new nodejs based approach and it seems to work fine! PTAL

academo (approved):
lgtm

.github/workflows/cd.yml
```diff
@@ -356,58 +361,69 @@ jobs:
...
-
-          else
-            echo "Invalid environment: ${ENVIRONMENT}"
-            exit 1
-          fi
+        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+        with:
+          script: |
+            // Helper to get input/env values
+            const getInput = (name, fallback = '') => {
```
tolzhabayev, line 368: why have a fallback if it's not used anywhere?

.github/workflows/cd.yml
```diff
@@ -356,58 +361,69 @@ jobs:
...
+            }
+
+            // Parse and filter environments
+            const allowedEnvironments = ['dev', 'ops', 'prod'];
+            let environments = [];
+            environments = ENVIRONMENT.split(',')
+              .map(e => e.trim())
+              .filter(e => allowedEnvironments.includes(e));
+
+            // Special case: 'prod' means all environments
```
tolzhabayev, line 406:
```suggestion
            // Special case: 'prod' means we deploy to all environments
```

.github/workflows/cd.yml
```diff
@@ -356,58 +361,69 @@ jobs:
...
+            if (environments.includes('prod')) {
+              environments = allowedEnvironments;
+            }
+
+            // Remove duplicates and sort
```
tolzhabayev, line 411: I think the "sort" part is missing, you could add `.sort()` to the Array.

`environments = Array.from(new Set(environments)).sort();`

.github/workflows/cd.yml
```diff
@@ -356,58 +361,69 @@ jobs:
...
+            const getInput = (name, fallback = '') => {
+              return process.env[name] !== undefined ? process.env[name] : fallback;
+            };
+
+            const ENVIRONMENT = getInput('ENVIRONMENT');
+            const DOCS_ONLY = getInput('DOCS_ONLY');
+            const HAS_BACKEND = getInput('HAS_BACKEND');
+
+            // Platforms matrix
+            let platforms;
```
tolzhabayev, line 377: more of a nitpick

```
let platforms = ['any'];

if (HAS_BACKEND === 'true') {
    platforms = ['linux', 'darwin', 'windows', 'any'];
}
```

.github/workflows/cd.yml
```diff
@@ -356,58 +361,69 @@ jobs:
...
+
+            // Docs only: skip plugin deployment
+            if (DOCS_ONLY === 'true') {
+              core.setOutput('environments', JSON.stringify([]));
+              core.setOutput('publish-docs', 'true');
+              return;
+            }
+
+            // Special case to skip deployment
+            if (!ENVIRONMENT || ENVIRONMENT === 'none') {
```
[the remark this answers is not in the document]
tolzhabayev: but then shouldn't this line be

https://github.com/grafana/plugin-ci-workflows/pull/135/files#diff-ea3ea8c9932adc7ba8161ceda844fedd43b006848ef1140c050cbd7ea0788a18R372

```
const ENVIRONMENT = getInput('ENVIRONMENT', 'none');
```
?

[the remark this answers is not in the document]
xnyo: No because we support passing a literal "none" string as a valid value for skipping the deployment (or an empty string).

```
const ENVIRONMENT = getInput('ENVIRONMENT', 'none');
```

This sets `ENVIRONMENT` to "none" if the `ENVIRONMENT` input is not provided. So we need extra logic to check for that "none" value

tolzhabayev (approved):

PR #140, opened 2025-07-02
Use GCS release artifacts to publish to catalogue

academo:
This PR fixes artifact consistency issues in plugin catalogue publishing by switching from "main" to "release" artifacts.

Previously, we used "main" artifacts  [generated in ci](https://github.com/grafana/plugin-ci-workflows/blob/0ff10ef11ee73912d45684a3820acdce88dd20ee/.github/workflows/ci.yml#L429). These artifacts were regenerated and reuploaded to GCS on each build, changing their checksums. Since gcom no longer allows republishing, this caused failures when plugins moved through dev → ops → prod environments, as the initial release contained different hashes than subsequent releases.

We already generate unique, immutable "release" artifacts that are never overwritten. This PR switches to using those instead.

## Implementation Details

the upload-to-gcs job runs parallel uploads for each platform's release artifacts. Each job generates URLs for all platform/architecture combinations (multiple URLs for plugins with backends, single "any" URL for frontend-only plugins).

GitHub matrix jobs overwrite each other's outputs, with the last job winning that's why we ensure all parallel jobs generate identical URL sets. This guarantees consistent output regardless of job completion order.

## Test with multiple platforms (darwin, linux, etc..)

A dev release https://github.com/grafana/plugins-drone-to-gha/actions/runs/16021344640

subsequent prod release: https://github.com/grafana/plugins-drone-to-gha/actions/runs/16021606758

notice in prod the ops publish to catalogue job failed for an unrelated reason.

## Test with a single `any` archive (frontend only)

Dev release: https://github.com/grafana/plugins-drone-to-gha/actions/runs/16025179331/job/45211456819


tolzhabayev (approved):
Looks good!

xnyo (approved):
LGTM, awesome work! 🎉

PR #166, opened 2025-07-21
feat: Add retry logic to publish request

### LINKED ISSUE #165: Enhancement: Add retry logic to gcom plugin publishing
originally open here https://github.com/grafana/grafana-community-team/issues/486

I deployed polystat to prod, which did all 3 environments, but dev gcom failed with this:

https://github.com/grafana/grafana-polystat-panel/actions/runs/16352995196/job/46204821501

```JSON
Publishing to https://grafana-dev.com/api
{
  "download": {
    "any": {
      "url": "https://redacted/grafana-polystat-panel/release/2.1.15/any/grafana-polystat-panel-2.1.15.zip"
    }
  },
  "url": "https://github.com/grafana/grafana-polystat-panel",
  "commit": "96d99e762eb2a5cdbab157baad9ab99cfac0f19b",
  "scopes": [
    "universal"
  ]
}
Response:
{
  "code": "InvalidArgument",
  "message": "Plugin with this version already exists. Please use a new version number.",
  "requestId": "4a3b32fa-65f1-4dcb-aa48-8c3eb362e019"
}
Plugin version already exists; IGNORE_CONFLICTS=true so treating as success
```

It did however succeed to publish, but the whole CD pipeline stopped and didn't proceed with the release in github.

https://grafana-dev.com/api/plugins/grafana-polystat-panel/versions/2.1.15

We could add some retry logic - if gcom fails, check if it really was accepted and call it a success)

oshirohugo:
Add retry logic to publish plugins request

Closes https://github.com/grafana/plugin-ci-workflows/issues/165

### DIFF
--- a/actions/plugins/publish/publish/publish.sh
+++ b/actions/plugins/publish/publish/publish.sh
@@ -78,7 +78,7 @@ curl_args=(
     "-H" "Accept: application/json"
     "-H" "User-Agent: github-actions-shared-workflows:/plugins/publish"
 )
-if [ "$has_iap" = true ]; then  
+if [ "$has_iap" = true ]; then
     if [ -z "$GCLOUD_AUTH_TOKEN" ]; then
         echo "GCLOUD_AUTH_TOKEN environment variable not set."
         exit 1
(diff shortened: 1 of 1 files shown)


briangann (commented):
This does add retry logic for the initial submit, but the issue I ran into was that the request was processed with no response.

Retrying here would end up with an error saying the plugin version already exists.

The logic needed is to:

1) Check if version exists

2) if it already exists, error out

3) Submit new version 

4) Check result code/response body
   a) If we get back a zero-length bod make another request to see if the version just submitted is in the catalog - and continue if it is. 
   b) If success, continue
   
5) If submit fails or #4a is false try to submit again.

xnyo:
I agree with @briangann , it's also more robust and clear to just check if the version already exists before attempting to publish. Note that we have to skip the POST request if `IGNORE_CONFLICTS` is true and the version already exists (this is required for prod publishing to work, because the version will most likely already exist in dev and ops)

PR #168, opened 2025-07-22
feat: add action to change plugin scope

oshirohugo:
Adds action to allow the scope(s) of a published plugin to be modified.

The workflow example example is added in this PR:
https://github.com/grafana/plugin-ci-workflows/pull/169


actions/plugins/publish/promote-plugin/action.yml
```diff
@@ -0,0 +1,43 @@
+name: Promote plugin
```
s4kh, line 1: Can we rename this to set plugin scope? Cause changing plugin scope is not actually promoting. It is more of a visibility thing. grafana_cloud: only on cloud, universal: all, grafana_cloud_org_{}: only on certain org.

actions/plugins/publish/change-plugin-scope/action.yml
```diff
@@ -0,0 +1,52 @@
...
+    required: true
+  gcloud-auth-token:
+    description: |
+      Google Cloud auth token for IAP access.
+      Required only for to dev or ops access.
+    required: false
+    default: ""
+  environment:
+    description: |
+      Environment where the plugin is published.
```
academo, line 15: we should clarify this. this is not exactly the environment where the plugin ispublished but the environment in which this workflow will change the scope.

actions/plugins/publish/change-plugin-scope/action.yml
```diff
@@ -0,0 +1,52 @@
...
+      Can be 'dev', 'ops', or 'prod'.
+    required: true
+  scopes:
+    description: |
+      Scope to promote the plugin to.
```
academo, line 20:
```suggestion
      Scope to apply to the plugin.
```
Let's not use the word "promote" (here and any other place in this workflow) we are not really promoting plugins here. Generally speaking this workflow let you change a plugin scope.

actions/plugins/publish/change-plugin-scope/action.yml
```diff
@@ -0,0 +1,52 @@
...
+    required: true
+  plugin-version:
+    description: |
+      Plugin version to promote.
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Promote plugin
```
academo, line 37:
```suggestion
    - name: Change plugin scope
```

academo (changes requested):
we should avoid calling any of this "promote" and instead "change plugin scope"

actions/plugins/publish/change-plugin-scope/action.yml
```diff
@@ -0,0 +1,52 @@
...
+  gcloud-auth-token:
+    description: |
+      Google Cloud auth token for IAP access.
+      Required only for to dev or ops access.
+    required: false
+    default: ""
+  environment:
+    description: |
+      Environment where the plugin is published.
+      Can be 'dev', 'ops', or 'prod'.
```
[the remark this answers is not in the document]
oshirohugo: I believe so.
The publish action accept several environments:
https://github.com/grafana/plugin-ci-workflows/blob/main/actions/plugins/publish/publish/action.yml#L23-L27
Therefore we can change the scope in all the environments where it's possible to publish a plugin.

xnyo (commented):
Great work! :raised_hands:  I think this is a great addition to the workflows! Some comments:

actions/plugins/publish/change-plugin-scope/change-plugin-scope.sh
```diff
@@ -0,0 +1,133 @@
...
+set +e
+echo $out | jq
+if [ $? -ne 0 ]; then
+    # Non-JSON output, print raw response
+    echo $out
+    exit 1
+fi
+
+# Determine if publish succeeded
+if [[ $(echo "$out" | jq -r '.scopes[]? | select(. == "universal")') == "universal" ]]; then
```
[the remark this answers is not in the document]
oshirohugo: That's true. This is a legacy error, since initially the action was changing the scope only to universal. I'll change it.

actions/plugins/publish/change-plugin-scope/change-plugin-scope.sh
```diff
@@ -0,0 +1,162 @@
+#!/bin/bash
```
academo, line 1: @oshirohugo might be too late to suggest this so if you don't feel like re-doing the work it's ok, but I always find these bash scripts harder to follow. 

Consider maybe migrating this script to nodejs and simply adding a setup-node step before this one.

xnyo (commented):
Some minor comments, mostly looks good to me !

xnyo (approved):
LGTM!

xnyo (approved):

academo (approved):

PR #169, opened 2025-07-23
docs: add example workflow for change scope plugins action

oshirohugo:
- Adding example/test workflow for the action introduced in https://github.com/grafana/plugin-ci-workflows/pull/168 (the branch will be changed after all the tests are done)
- Improving docs for clarity


.github/workflows/promote-plugin-scope.yml
```diff
@@ -0,0 +1,123 @@
...
+# Promote plugin scope to universal.
+# It uses action grafana/plugin-ci-workflows/actions/plugins/publish/promote-plugin
+# to change plugin scope to universal.
+# The following inputs should be provided:
+# - environment: Target environment (dev, ops, prod)
+# - plugin_id: Plugin ID to promote
+# - plugin_version: Plugin version to promote
+#
+
+name: Plugins - Promote to Universal Scope
```
tolzhabayev, line 11: I wonder if we want to make this a generic action - like "Set plugin's scope" and make scope and input field which default to universal. wdyt?

s4kh (approved):

.github/workflows/change-plugin-scope.yml
xnyo, line 1: In my opinion, this workflow is better suites as a an "Extra CI helper", since it doesn't have to be part of the base workflows for CI/CD to work. I would put this in `examples/extra` folder instead

.github/workflows/change-plugin-scope.yml
```diff
@@ -0,0 +1,113 @@
...
+    permissions:
+      id-token: write
+      contents: read
+
+    env:
+      # Set defaults for PR triggers, use inputs for manual dispatch
+      ENVIRONMENT: ${{ inputs.environment || 'dev' }}
+      PLUGIN_ID: ${{ inputs.plugin_id || 'hugooshiro-dummy-datasource' }}
+      PLUGIN_VERSION: ${{ inputs.plugin_version || '1.0.5' }}
+      SCOPES: ${{ inputs.scopes || 'universal' }}
```
xnyo, line 57: Nit: I think it's easier if we move the default values up in the workflow file, in an workflow-level `env` section.

This is a pattern I also used in the example workflows for the base CI/CD (for example, see [here](https://github.com/grafana/plugin-ci-workflows/blob/5376f13336852f0f22ae7ad03a4734a6d997a6f1/examples/base/simple/publish.yaml#L31-L36) or [here](https://github.com/grafana/plugin-ci-workflows/blob/5376f13336852f0f22ae7ad03a4734a6d997a6f1/examples/base/provisioned-plugin-auto-cd/publish.yaml#L31-L37)), so all the configurable options are in the same place, along with a TODO comment which tells the user it should be addressed before installing the workflow.

Suggested changes:

```yaml
on:
  # ...

env:
  # Default input values, can be overwritten via workflow_dispatch inputs
  ENVIRONMENT: dev
  PLUGIN_ID: hugooshiro-dummy-datasource  # TODO: change with your plugin's slug
  SCOPES: universal  # TODO: change with an appropriate value

jobs:
  # ...
```

And then here:

```yaml
jobs:
  promote:
    # ...
    env:
      # Set defaults for PR triggers, use inputs for manual dispatch
      ENVIRONMENT: ${{ inputs.environment || env.ENVIRONMENT }}
      PLUGIN_ID: ${{ inputs.plugin_id || env.PLUGIN_ID }}
      PLUGIN_VERSION: ${{ inputs.plugin_version }}
      SCOPES: ${{ inputs.scopes || env.SCOPES }}
```

I think this approach _should_ work, but I'm not 100% sure since we are using `env.*` (which should be the workflow-level environment) in a job-level environment section, so it may behave differently

.github/workflows/change-plugin-scope.yml
```diff
@@ -0,0 +1,113 @@
...
+
+    permissions:
+      id-token: write
+      contents: read
+
+    env:
+      # Set defaults for PR triggers, use inputs for manual dispatch
+      ENVIRONMENT: ${{ inputs.environment || 'dev' }}
+      PLUGIN_ID: ${{ inputs.plugin_id || 'hugooshiro-dummy-datasource' }}
+      PLUGIN_VERSION: ${{ inputs.plugin_version || '1.0.5' }}
```
xnyo, line 56: Maybe I'm missing something, but I think the version should be required and always provided explicitly as a input?

.github/workflows/change-plugin-scope.yml
```diff
@@ -0,0 +1,113 @@
...
+        required: true
+        default: 'dev'
+        type: choice
+        options:
+          - dev
+          - ops
+          - prod
+      plugin_id:
+        description: 'Plugin ID to change scope'
+        required: true
```
xnyo, line 29: This is not required anymore since the workflow has logic to handle a default value when the input is empty

```suggestion
        required: false
```

.github/workflows/change-plugin-scope.yml
```diff
@@ -0,0 +1,113 @@
...
+        description: 'Plugin version to change scope'
+        required: true
+        type: string
+      scopes:
+        description: |
+          Scope to change the plugin to.
+          Can be 'universal', 'grafana_cloud', 'grafana_cloud_org' or 'grafana_cloud_instance'.
+          Multiple values can be provided separated by commas.
+        required: false
+        default: 'universal'
```
xnyo, line 41: This should default to an empty string for the default value logic to work:

```yaml
SCOPES: ${{ inputs.scopes || 'universal' }}
```

```suggestion
```

xnyo (commented):
Great work! :raised_hands:  Some comments:

.github/workflows/change-plugin-scope.yml
oshirohugo: Reverencing the env in the second part didn't work. So I change it to set the plugin-id through environment variables as you recommended, since it will be a value that will only change from repository to repository. And I set all the other inputs as required, so no need for default value.

xnyo (approved):
Awesome work! I haven't tested it but the code LGTM! :rocket:

PR #185, opened 2025-07-31
feat: add optional node and npm dependency caching

sd2k:
Similar to https://github.com/grafana/plugin-ci-workflows/pull/181 but for node and npm. This has to be manually enabled
because we don't know the user's package manager.

### DIFF
--- a/actions/plugins/setup/action.yml
+++ b/actions/plugins/setup/action.yml
@@ -19,6 +19,14 @@ inputs:
     description: Node.js version file to use.
     required: false
     default: ""
+  node-cache:
+    description: Used to specify a package manager for caching in the default directory. Supported values npm, yarn, pnpm (https://github.com/actions/setup-node#caching-global-packages-data)
+    required: false
+    default: ""
+  node-cache-dependency-path:
+    description: Used to specify the path to a dependency file - package-lock.json, yarn.lock, etc. Supports wildcards or a list of file names for caching multiple dependencies (https://github.com/actions/setup-node#caching-global-packages-data)
+    required: false
+    default: ""
 
 runs:
   using: composite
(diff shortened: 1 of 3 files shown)


xnyo (commented):
Great work! :raised_hands: I think this looks good and can be merged, but would love an opinion from @jackw, as he was looking into adding a shared action to detect the package manager being used in a plugin's repo.

Once that's introduced, we can make the two arguments optional and, when they're not provided, the package manager will be auto-detected. However we probably still need an input to disable caching if necessary.

I think we can merge this for now, but we may have to change/deprecate some inputs in the future. @jackw wdyt? maybe we should wait until we can auto-detect the package manager?

tolzhabayev:
fixes https://github.com/grafana/plugin-ci-workflows/issues/194

actions/plugins/setup/action.yml
```diff
@@ -23,6 +31,8 @@ runs:
       uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
       with:
         node-version: "${{ inputs.node-version }}"
+        cache: "${{ inputs.node-cache }}"
```
tolzhabayev, line 34: do we know what happens when those inputs are empty?

sd2k: [Empty string is the default](https://github.com/actions/setup-node#usage) so it should work fine.

sd2k:
I've updated this to remove the `node-cache` input, which expected the package manager name, and use the existing `package-manager` input instead. Now the only new thing in the external ci/cd.yml files is the optional `node-cache-dependency-path` input which should take package.lock, yarn.lock etc.

I imagine we use @jackw's PR to automatically populate a sensible default here but the underlying `setup-node` action [allows more customisation](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-data) (e.g. for multiple package.lock files) so I don't think it's crazy to keep this option exposed - WDYT?

PR #190, opened 2025-08-01
docs: Add README to each base example and add script to generate root README

xnyo:
Copied each example's README.md into its own directory. This adds a README visible in GitHub when linking a specific example.

Adds a script to generate the root folder's README.md file, by combining all the README.md files for each example.


tolzhabayev (approved):
Feels like the go run genreadme.go should be automatically run as a gh actions job in PRs when one of the relevant files changes or you will drift quickly. 

But maybe for the future.

PR #196, opened 2025-08-01
fix(publish): Fix secrets not being passed to Playwright when publishing

xnyo:
Fixes an issue when using Playwright secrets where the secrets are not fetched from Vault when the CI is triggered from CD (via workflow_dispatch).

This was caused because the expression to determine if secrets should be passed was:

```
github.event.pull_request.head.repo.full_name == github.repository && github.actor != 'dependabot[bot]'
```

But for CD triggers, the event is NOT a pull_request, but rather a workflow_dispatch, so it always evaluates to false and causes the secrets to be empty.

Example failure in the x-ray-datasource: https://github.com/grafana/x-ray-datasource/actions/runs/16673945572

This PR fixes the issues and refactors the logic to determine if secrets should be fetched or not. Rather than using those complex expressions, it adds a new `workflow-context` step to the `test-and-build` job, which uses a small node js snippet to export some useful information about the workflow run (is fork, is trusted, is bot, etc). Those can then be used to more easily trigger bits of the workflow depending on whether the PR comes from a trusted source or not.

**Keep in mind WIF will still prevent accessing Vault in case the PR comes from outside Grafana Labs. This logic that skips/omits some secrets is only to allow the workflow to pass. Otherwise it would fail when trying fetch secrets from Vault (it fails with a 403).**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -252,7 +252,7 @@ jobs:
               github.event_name === 'workflow_dispatch' || (isPR && !isForkPR)
             ) && !isBot;
 
-            return { isTrusted, isForkPR };
+            return { isTrusted, isForkPR, isBot };
 
       - name: Setup
         uses: grafana/plugin-ci-workflows/actions/plugins/setup@main # zizmor: ignore[unpinned-uses]
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -240,7 +240,7 @@ jobs:
         id: workflow-context
         with:
           script: |
-            console.log('input context:', JSON.stringify(github));
+            console.log('input context:', JSON.stringify(context));
 
             const bots = [
               'dependabot[bot]',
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -228,13 +226,34 @@ jobs:
...
           ref: ${{ inputs.branch }}
           persist-credentials: false
 
+      - name: Determine workflow context
+        uses: actions/github-script@60a0d83039c74a4aee543508d2ffcb1c3799cdea # v7.0.1
+        id: workflow-context
+        with:
+          script: |
+            const bots = [
+              'dependabot[bot]'
```
tolzhabayev, line 244:
```suggestion
              'dependabot[bot]',
              'renovate[bot]',
              'github-actions[bot]',
              'renovate-sh-app[bot]'
```

tolzhabayev (approved):
awesome catch, lgtm - small bot list adjustment

PR #199, opened 2025-08-04
feat(publish): Sanity check ZIP files before publishing to catalog

xnyo:
Adds a new re-usable action: `plugins/publish/check-artifacts`.

This action runs the following sanity checks on a list of plugin ZIP files:

- Checks that there's a root folder with the plugin's id as its name
- Checks that MANIFEST.txt is present (plugin is signed)
- Checks that plugin.json is present

This action is run before publishing a plugin, to ensure that broken plugins aren't published to the catalog

Example run: https://github.com/grafana/plugins-drone-to-gha/actions/runs/16773289220/job/47493770200

### DIFF
--- a/actions/plugins/publish/check-artifacts/check-artifacts.sh
+++ b/actions/plugins/publish/check-artifacts/check-artifacts.sh
@@ -44,12 +44,12 @@ validate_plugin_zip() {
 }
 
 process_plugin_zips() {
-    local json_array="$1"
+    local urls_array="$1"
     local plugin_id="$2"
     local failed_count=0
 
     # Parse JSON array and process each URL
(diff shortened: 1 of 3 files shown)


actions/plugins/publish/check-artifacts/check-artifacts.sh
```diff
@@ -0,0 +1,84 @@
...
+        return 0
+    fi
+}
+
+# Main script execution
+if [[ $# -ne 2 ]]; then
+    echo "Usage: $0 <json_array_of_urls_or_single_url> <plugin_id>"
+    exit 1
+fi
+json_array="$1"
```
academo, line 73: nit: rename this to urls_array (instead of naming it after its type, name it for what it contains)

academo (approved):

PR #202, opened 2025-08-04
fix(cd): github draft optional, omit commits

L2D2Grafana:
- Update cd.yml with an optional configuration for github-draft-release. It's set to true by default for backwards compatibility. This flexibility will be nice because some of the other changelog libraries do the draft release for you (ex: release please). 
- Allow generate-changelog to omit commits


xnyo (approved):
LGTM! :rocket:

academo (approved):

PR #204, opened 2025-08-05
feat(ci)!: use pm-detect to understand nodejs package manager

### LINKED ISSUE #194: Add cache to setup-node step
## Why?

Speed workflows up by adding cache configuration to 
https://github.com/grafana/plugin-ci-workflows/blob/47f6682ca3c59a62aa4ead4269bb1f06484dd7dc/actions/plugins/setup/action.yml#L23

We already have an input for which package manager is used https://github.com/grafana/plugin-ci-workflows/blob/47f6682ca3c59a62aa4ead4269bb1f06484dd7dc/.github/workflows/ci.yml#L48

Decide on - make the cache an opt-in or opt-out?

jackw:
We've been using pm.sh files across multiple workflow repos to try to understand which package manager plugins are using. However with the pm-detect workflow in plugins-actions we can remove the pm.sh files and the shared workflow will figure out a plugins node package manager for us and return that tools commands.

See grafana/plugin-actions/pull/112

fixes: #194 
fixes: #185

related: #405


mckn (approved):
LGTM!

.github/workflows/cd.yml
```diff
@@ -71,15 +71,10 @@ on:
...
-        description: Directory of the plugin, if not in the root of the repository. If provided, package-manager must also be provided.
+        description: Directory of the plugin, if not in the root of the repository.
         type: string
         required: false
         default: .
-      package-manager:
-        description: The package manager to use.
-        type: string
-        required: false
-        default: ""
```
xnyo: This is a breaking change. Please rename the PR to:

```
feat(ci)!: use pm-detect to understand nodejs package manager
```

So release-please picks it up as a breaking change and releases a new major version

.github/workflows/playwright.yml
```diff
@@ -109,11 +109,23 @@ jobs:
...
 
+      - name: Detect package manager
+        id: package-manager
+        uses: grafana/plugin-actions/package-manager-detect@package-manager-detect/v1.0.0
+        with:
+          working-directory: ${{ inputs.plugin-directory }}
+
+      - name: Install pnpm
+        if: steps.package-manager.outputs.name == 'pnpm'
+        uses: pnpm/action-setup@a7487c7e89a18df4991f7f222e4898a00d66ddda # v4.1.0
```
xnyo: Nit: We can probably refactor this to avoid repeating it both here and in the setup action (maybe calling the setup action in this step, and add somehing like a `frontend-only` input to the setup action to skip the backend setup steps), but we can leave it for a later follow-up PR

xnyo (commented):
Code changes LGTM! Only a couple of comments. However act tests are failing, taking a look at them 👀

xnyo (approved):
LGTM, awesome work! 🥇

Issue #97, opened 2025-05-16
Allow customizing runner size


xnyo:
Slack thread: https://raintank-corp.slack.com/archives/C0671T2J5CH/p1747311975354289

Add an input to be able to specify which runner (or runner size) is used for building the plugin and running e2e tests.

Currently, it defaults to `ubuntu-latest-8-cores`. However, those runners have limits and we don't really need such big runner for small plugins. The default should be the smaller `ubuntu-latest` (which is also free on public repos).

We should also use `github-hosted-ubuntu-x64-large` instead of `ubuntu-latest-cores`.

xnyo:
Related to https://github.com/grafana/plugin-ci-workflows/issues/213

PR #217, opened 2025-08-07
feat: Switch runners for exported workflows to self-hosted ones

xnyo:
Switches the runners for the ci, cd and extra workflows to self-hosted ones.

Part of #213

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -217,7 +217,7 @@ env:
 jobs:
   test-and-build:
     name: Test and build plugin
-    runs-on: ubuntu-x64-large
+    runs-on: ubuntu-x64-large-io
 
     outputs:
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -217,7 +217,7 @@ env:
 jobs:
   test-and-build:
     name: Test and build plugin
-    runs-on: ubuntu-latest-8-cores
+    runs-on: ubuntu-x64-large
```
xnyo, line 220: This is "large" (as well as the runner used for Playwright e2e tests) for backwards compatibility (self-hosted large has 8 cores, vs the default 4). There's an issue to allow customizing the size of the runner (#97)

andresmgot (approved):
LGTM

PR #219, opened 2025-08-07
fix: pass node-version down from ci to playwright

oshirohugo:
- Pass node-version down from ci.yml to playwright.yml
- Remove the default node version and let the node version fallback for the nvmrc when as it was before.


briangann (approved):
LGTM!

wbrowne (approved):

PR #237, opened 2025-08-14
feat: add option to disable docs publishing alltogether

academo:


.github/workflows/cd.yml
```diff
@@ -828,6 +835,7 @@ jobs:
       ${{
         (needs.ci.outputs.has-docs == 'true')
         && (needs.define-variables.outputs.publish-docs == 'true')
+        && !inputs.disable-docs-publishing
```
academo, line 838: overkill since the define variables covers it but here to make sure things the logic is explicit.

andresmgot (approved):

PR #240, opened 2025-08-18
feat: add publish to catalog as pending option in ci and publish action

### LINKED ISSUE #239: Workflow input for "Hiding from the catalog" for plugins that want to target cloud users but also be only installed via provisioning
## Why?
It seems like when a new version of a plugin gets published, it automatically sets the `status` to `active`, even if it was `pending` before.

## What
Add workflow input that gets passed in and then carried over to the body of publishing request setting `pending:true` in the body of the rquest - see https://admin.grafana.com/api-docs#/plugins/postPlugins

academo:
closes https://github.com/grafana/plugin-ci-workflows/issues/239

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -249,6 +249,12 @@ on:
         type: string
         required: false
         default: ""
+      publish-to-catalog-as-pending:
+        description: |
+          If true, the plugin will be published as pending in the plugins catalog.
+          Default is false.
+        default: false
+        type: boolean
 
       # Options for deploying PRs. Those values should come from the PR event and should not be set manually.
       branch:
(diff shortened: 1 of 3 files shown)


s4kh (approved):

xnyo (commented):
Code looks good. I will test this shortly. In the meantime, can you also update the docs in EngHub please? This part here:

https://enghub.grafana-ops.net/docs/default/component/grafana-plugins-platform/plugins-ci-github-actions/040-plugins-cd/#special-case-cloud-only-plugins-that-want-to-be-rolled-out-selectively

Maybe also the docs in SSS need to be updated?

tolzhabayev (approved):

xnyo (approved):
LGTM

PR #253, opened 2025-08-25
feat: add step to cache playwright

### LINKED ISSUE #182: PlayWright setup times improvements
## Why?
It takes a while to get playwright installed in github action steps, we are round about at 7 minutes right now and half of it goes to playwright install step.

Example runs: 
- https://github.com/grafana/grafana-assistant-app/actions/runs/16648652844/job/47115830755

## What
Take a look at 
- using a pre-backend image from microsoft - see https://github.com/microsoft/playwright/issues/7249
- or try to cache it manually like this https://github.com/hirasso/thumbhash-custom-element/blob/main/.github/actions/playwright/action.yml

oshirohugo:
### What is the purpose of this PR
Add step to cache playwright and improve playwright setup times.

### Note for reviewers
The cache approach was chosen over the pre-backend image from microsoft because using a specific image brought some complexities to run the plugin environment using `docker compose`. We would have to run docker inside docker and make the plugin environment provided by docker compose be aware of the network where the `pre-backend image` was running, which is only known during the action runtime. This information would have to be added to the plugin `docker-compose.yml` before running `docker compose up`

**Run example**
https://github.com/grafana/grafana-pluginsplatformprovisioned-app/actions/runs/17218721506/job/48849154440

fixes https://github.com/grafana/plugin-ci-workflows/issues/182

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -450,7 +450,7 @@ jobs:
 
   playwright:
     name: Playwright E2E tests
-    uses: grafana/plugin-ci-workflows/.github/workflows/playwright.yml@cache-playwright # zizmor: ignore[unpinned-uses]
+    uses: grafana/plugin-ci-workflows/.github/workflows/playwright.yml@main # zizmor: ignore[unpinned-uses]
     if: ${{ inputs.run-playwright == true }}
     needs:
       - test-and-build
@@ -472,7 +472,7 @@ jobs:
 
   playwright-docker:
     name: Plugins - Dockerized Playwright E2E tests
-    uses: grafana/plugin-ci-workflows/.github/workflows/playwright-docker.yml@cache-playwright # zizmor: ignore[unpinned-uses]
+    uses: grafana/plugin-ci-workflows/.github/workflows/playwright-docker.yml@main # zizmor: ignore[unpinned-uses]
     if: ${{ inputs.run-playwright-docker == true }}
     needs:
       - test-and-build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -444,7 +444,7 @@ jobs:
 
   playwright:
     name: Playwright E2E tests
-    uses: grafana/plugin-ci-workflows/.github/workflows/playwright.yml@main # zizmor: ignore[unpinned-uses]
+    uses: grafana/plugin-ci-workflows/.github/workflows/playwright.yml@cache-playwright # zizmor: ignore[unpinned-uses]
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -444,7 +444,7 @@ jobs:
 
   playwright:
     name: Playwright E2E tests
-    uses: grafana/plugin-ci-workflows/.github/workflows/playwright.yml@main # zizmor: ignore[unpinned-uses]
+    uses: grafana/plugin-ci-workflows/.github/workflows/playwright.yml@cache-playwright # zizmor: ignore[unpinned-uses]
```
oshirohugo, line 447: this will be changed to `main` once the PR is accepted.

.github/workflows/playwright.yml
```diff
@@ -141,6 +141,36 @@ jobs:
         shell: bash
         working-directory: ${{ inputs.plugin-directory }}
 
+      - name: Get Playwright version
+        uses: actions/github-script@v7
```
xnyo, line 145: This should be pinned to the exact hash

.github/workflows/playwright.yml
```diff
@@ -141,6 +141,36 @@ jobs:
...
+            if (version) {
+              core.exportVariable("PLAYWRIGHT_VERSION", version);
+              core.setOutput("version", version);
+            } else core.setFailed("Couldn't get Playwright version");
+        env:
+          PLUGIN_DIRECTORY: ${{ inputs.plugin-directory }}
+
+      - name: Cache Playwright
+        id: cache
+        uses: actions/cache@v4
```
xnyo, line 169: This should be pinned to the exact hash

.github/workflows/playwright.yml
```diff
@@ -141,6 +141,36 @@ jobs:
...
+            let version = "";
+            try {
+              version = require("@playwright/test/package.json").version;
+            } catch (error) {
+              console.log(error.message);
+            }
+            console.debug("Version:", version);
+            if (version) {
+              core.exportVariable("PLAYWRIGHT_VERSION", version);
+              core.setOutput("version", version);
```
xnyo, line 162: Do we need to export the variable for future steps, or does `core.setOutput` suffice? (and then we can use it in the next step as `${{ steps.version.outputs.version }}` rather than `${{ env.PLAYWRIGHT_VERSION }}`)

xnyo (commented):
Great work! Some minor comments/suggestions:

xnyo (approved):
LGTM! Awesome work! :raised_hands: Remember to switch back to `@main` branch before merging

PR #270, opened 2025-09-02
ci: Add workflow for switching to rolling release after merging tags

xnyo:


.github/workflows/restore-rolling-release.yml
```diff
@@ -0,0 +1,26 @@
...
+jobs:
+  switch-references:
+    name: Switch references
+    runs-on: ubuntu-x64-small
+    steps:
+      - name: Checkout main branch
+        uses: actions/checkout@08eba0b27e820071cde6df949e0beb9ba4906955 # v4.3.0
+        with:
+          ref: main
+          persist-credentials: false
```
tolzhabayev, line 21: you have `persist-credentials: false` - don't you need the token / credentials to be able to push the changed references?

PR #280, opened 2025-09-05
fix: add prod-canary

s4kh:
Part of https://github.com/grafana/grafana-community-team/issues/544

We ONLY publish `status:pending` and `version+commit hash` to the catalog when env contains `prod-canary` but not `prod` - user specifically requests prod-canary deployment. When `prod` is in the environment, it follows the usual steps

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -203,6 +203,7 @@ on:
           - `none` (or empty string): Skip deployment (run only CI)
           - `dev`: Deploy to dev catalog
           - `ops`: Deploy to ops catalog
+          - `prod-canary`: Deploy to subset of prod instances - free and on instant wave
           - `prod`: Deploy to dev AND ops AND prod
           - A comma separated combination of the values above. E.g.: `dev,ops`
 
@@ -330,9 +331,9 @@ jobs:
             echo "plugin_version_suffix=${INPUT_PLUGIN_VERSION_SUFFIX}" >> "$GITHUB_OUTPUT"
           else
             # Default behavior:
-            # - Use the commit SHA as the version suffix when not on main
+            # - Add commit hash for prod-canary only deployments OR non-main branches
             # - Otherwise, do not set the suffix
-            if [ "${INPUT_BRANCH}" != 'main' ]; then
+            if [[ ("${ENVIRONMENT}" == *"prod-canary"* && "${ENVIRONMENT}" != *"prod"*) || "${INPUT_BRANCH}" != 'main' ]]; then
               echo "plugin_version_suffix=${COMMIT_SHA}" >> "$GITHUB_OUTPUT"
             else
               echo "plugin_version_suffix=" >> "$GITHUB_OUTPUT"
@@ -341,6 +342,7 @@ jobs:
         env:
           INPUT_PLUGIN_VERSION_SUFFIX: ${{ inputs.plugin-version-suffix }}
           INPUT_BRANCH: ${{ inputs.branch }}
+          ENVIRONMENT: ${{ inputs.environment }}
           COMMIT_SHA: ${{ steps.checkout-specified-branch.outputs.commit }}
         shell: bash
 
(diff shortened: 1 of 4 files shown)


.github/workflows/cd.yml
```diff
@@ -453,7 +453,7 @@ jobs:
             }
 
             // Parse and filter environments
-            const allowedEnvironments = ['dev', 'ops', 'prod'];
+            const allowedEnvironments = ['dev', 'ops', 'prod-canary', 'prod'];
```
s4kh, line 456: @xnyo We are adding prod-canary only to be included when prod is passed, right?

s4kh: When we pass target, it publishes the plugin to that env's catalog and we don't have prod-canary catalog(it is prod), not exactly sure how that works?

s4kh: When target is prod-canary:
1. Publish to prod catalog - Would this cause any errors?
2. Trigger argo workflow - this will work fine
3. Should we publish the docs?
After deploying to prod-canary, it is called with prod
1. Publishing again to prod - Would this cause errors? Duplicate version error?

.github/workflows/cd.yml
```diff
@@ -618,10 +618,16 @@ jobs:
       - name: Prepare parameters
         id: argo-parameters
         run: |
+          # Add prod-canary when prod is present in the environment
+          modified_environment="${ENVIRONMENT}"
+          if [[ "${ENVIRONMENT}" == *"prod"* && "${ENVIRONMENT}" != *"prod-canary"* ]]; then
```
s4kh, line 623: @xnyo for now, should we just trigger argo-workflow when prod is present in the environment so it updates the prod-canary `_catalog_version`.

.github/workflows/cd.yml
```diff
@@ -453,15 +455,16 @@ jobs:
...
             let environments = [];
             environments = ENVIRONMENT.split(',')
               .map(e => e.trim())
               .filter(e => allowedEnvironments.includes(e));
 
             // Special case: 'prod' means we deploy to all environments
             if (environments.includes('prod')) {
-              environments = allowedEnvironments;
+              // no need to deploy to prod-canary because it is a subset of prod
+              environments = allowedEnvironments.filter(e => e !== 'prod-canary');
```
s4kh, line 467: When we have prod → we are publishing to all the catalogs, no need to publish twice for prod. Argo workflow will take care of prod-canary trigger in the [code](https://github.com/grafana/deployment_tools/blob/master/ksonnet/environments/grafana-plugins-cd/main.jsonnet#L41)

.github/workflows/cd.yml
```diff
@@ -203,6 +203,7 @@ on:
           - `none` (or empty string): Skip deployment (run only CI)
           - `dev`: Deploy to dev catalog
           - `ops`: Deploy to ops catalog
+          - `prod-canary`: Deploy to subset of prod instances - free and on instant wave
           - `prod`: Deploy to dev AND ops AND prod
```
xnyo, line 207: Nit:

```suggestion
          - `prod`: Deploy to dev AND ops AND prod-canary AND prod
```

xnyo (commented):
Great work! :1st_place_medal: I have left some minor comments, but the changes look good in general. We should also find a way to test this, we don't really have the testing plugin deployed to prod-canary or prod so I'll have to figure out something first :thinking:

s4kh:
@xnyo 

> We should also find a way to test this, we don't really have the testing plugin deployed to prod-canary or prod so I'll have to figure out something first

I have tested the plugin-version suffix adding part last Friday using `adaptivelogs-app` with Darren but failed at the Upload GCS(tags) due to service accounts only run on `main` branch workflows. Somehow we release this as a 2.0.1-beta version and test with plugin whoever is publishing a new version?

xnyo (approved):
LGTM! :rocket:  I guess the easiest way is merge this and test with a real plugin using `@main` directly. If there are any issues, users can roll back to `@ci-cd-workflows/v2.0.0`

PR #285, opened 2025-09-09
chore: add zizmor config and ignore "unpinned-uses" for plugin-ci-workflows references

xnyo:
Adds a repo-level zizmor config to avoid having to specify `zizmor: ignore[unpinned-uses]` for all references of plugin-ci-workflows actions/workflows


tolzhabayev (approved):

tolzhabayev:
btw. shouldn't this config actually work https://github.com/grafana/shared-workflows/blob/main/.github/zizmor.yml#L9?

PR #288, opened 2025-09-09
chore: updating renovate config

tolzhabayev:
- `.github/renovate.json5` is not supported in self hosted renovate runner so it is currently just being ignored
- Adding minimumReleaseAge


xnyo (approved):
Awesome!

PR #291, opened 2025-09-10
chore(deps)!: Bump default Node.js version used to build plugin from 20 to 22 (Active LTS)

xnyo:
Bumps the version used to build the plugin from Node.js 20 to Node.js 22.

Node.js 22 is the new current active LTS, and Node.js 20 is in maintenance mode.

See here for details: https://nodejs.org/en/about/previous-releases


tolzhabayev (approved):
to anyone finding this PR - you can set the node version [via inputs](https://github.com/grafana/plugin-ci-workflows/blob/1f78e66915b8f734504e24299b5eb94126a1949e/.github/workflows/ci.yml#L25). So this is only the default fallback version, you are not limited to only use this one.

PR #300, opened 2025-09-18
feat(workflows): fail ci for failed signs on ops and prod

### LINKED ISSUE #298: CI Pipeline should fail if a plugin fails to sign for ops and prod environments

oshirohugo:
## What does this PR do?
Add the following inputs to CD:
- `force-allow-unsigned`

Add the following inputs to CI:
- `environment` - it will be passed down from CD, and if present, if will not allow unsigned plugins to be packaged if the environment is not `dev`. If it's `dev` or if it's not used it will keep following the same logic as before which is controlled by `isTrusted`
- `force-allow-unsigned` - it will be passed down from CD, and it will force allow unsigned plugins to be package, regardless of the environment or the value of `isTrusted`

## Note for reviewers:
It also worth to mention the currently the CI already fails when there is an error in the plugin sign up command:
https://github.com/grafana/plugin-ci-workflows/blob/main/actions/plugins/package/package.sh#L10
if this step returns 1, the CI will fail.

But currently it's possible to skip this step for any environment as long as `isTrusted` is false:
https://github.com/grafana/plugin-ci-workflows/blob/main/.github/workflows/ci.yml#L273

closes https://github.com/grafana/plugin-ci-workflows/issues/298

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -367,7 +367,7 @@ jobs:
 
   ci:
     name: CI
-    uses: grafana/plugin-ci-workflows/.github/workflows/ci.yml@main
+    uses: grafana/plugin-ci-workflows/.github/workflows/ci.yml@fail-ci-for-failed-signs-on-ops-and-prod
     needs:
       - setup
     with:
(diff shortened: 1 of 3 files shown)


xnyo (commented):
I agree with Esteban regarding changing the input name. I also find it a bit confusing. The rest of the changes look good to me! :raised_hands:

academo (approved):

xnyo (approved):
LGTM!

PR #315, opened 2025-09-30
feat(ci): add job to check for workflows references in pr-checks.yml

### LINKED ISSUE #297: Add status check to ensure all references point to `@main`
When testing changes in plugin-ci-workflows in one of the testing repos, we almost always have to switch all references from `@main` to `@pr-branch`. These changes have to be reverted before merging the PR.

We should add a status check to ensure that all references point to `@main` rather than the PR's branch, as a sanity check.

xnyo:
Adds a PR check workflow to check if all referenced workflows are pointing to `@main`, to ensure there are no leftovers from testing. This workflow is internal to plugin-ci-workflows only.

Fixes #297.

Example:
immagine


academo (approved):

.github/workflows/pr-checks-workflow-references.yml
```diff
@@ -0,0 +1,67 @@
...
+            .github/workflows/ci.yml
+            .github/workflows/cd.yml
+            .github/workflows/playwright.yml
+            .github/workflows/playwright-docker.yml
+            actions/plugins/**
+
+      - name: Check drift
+        if: steps.switch-references.outputs.changed == 'true'
+        run: |
+          git diff
```
tolzhabayev, line 43: this diff (line 43) is a leftover from testing?

xnyo: I wanted to leave it so the raw git diff is still available in the GitHub actions logs, just in case there are any issues with the awk script below for the fancy GitHub actions annotations

tolzhabayev (approved):

Issue #296, opened 2025-09-10
Investigate ways to improve testing changes to the workflows


xnyo:
Currently, testing changes in plugin-ci-workflows is not straightforward, and there aren't any automated checks in PRs to see if the workflows still function properly with the changes made in a PR.

This can be particularly useful for testing version bumps in renovate PRs: we often run a PR's branch in the testing repository manually before submitting the PR, but we don't usually do that for the automated PRs opened by renovate. For example, https://github.com/grafana/plugin-ci-workflows/pull/273 introduced a breaking change that broke CI, but it was reported only by the slo team after the PR was merged.

Some ideas, not sure how feasible they are:

---

**A) GHA workflows for testing plugin-ci-workflows**

We can have some workflows (`on: pull_request`) in plugin-ci-workflows that:
- Clone one or more test plugin into the GHA runner (like https://github.com/grafana/plugins-drone-to-gha for a plugin with a backend, https://github.com/grafana/clock-panel for a frontend-only one, etc). Alternatively we could also have a folder with various testing plugins embedded in the plugin-ci-workflows repository. This can be handy for testing the various js package managers for example
- Run plugin-ci-workflows's CI for those plugins, using the branch of the current PR:
  - Run the switch references action for the current repository, so all `@main` references are updated to `@pr-branch`. Not sure if this would work? Some GHA caching may get in the way?
  - Run CI via `uses: ./.github/workflows/ci.yml`, so locally instead of remotely. Not sure if this would work? GHA behaves differently when referencing local vs remote workflows. I remember trying using locally checked out workflows in GHA before and it didn't work as expected
- Not sure how to confusing it would be to test CD in this way
- WIF might get in the way for GCS/publishing steps and may need to be updated. Otherwise the workflow will always fail because access is not allowed from plugin-ci-workflows non-main branches

---

**B) Run CI periodically (cron) in the testing repos**

We can run CI/CD periodically in the testing repos:
- https://github.com/grafana/plugins-drone-to-gha
- https://github.com/grafana/grafana-pluginsplatformprovisioned-app

And report back to Slack if it's failing. Those repos should use `@main` (rolling release) rather than a pinned version so we test the latest changes. However, these tests however would only kick in after merging a PR, and we cannot define status checks on this. Not a huge deal since we have proper semantic versions now, we have time to test before cutting a release.

---

**C) Call the workflow in the testing repos?**

[...]

PR #339, opened 2025-10-07
feat: add smoke tests for the shared ci workflow

xnyo:
Adds a system for smoke testing the CI workflow. Implements solution "A" of #296.

This PR introduces:

- A `tests` folder, which will contain various testing setups for testing plugin-ci-workflows
  - Insite of it, two new empty plugins scaffolded with create-plugin: a panel (frontend-only, with npm) and a datasource with a backend
- A new GHA workflow run for PRs and pushes to main that will run the CI workflow for all the plugins inside the `tests` folder, using the version of plugin-ci-workflows checked out locally (so the PR's version of the content, or the current `main` branch). This workflow runs CI for each test plugin. Playwright is disabled (for now) as well as steps that require Vault access (for security reasons, access through WIF to Vault and GCS is not granted)
- Adds some inputs to the CI workflow to allow running multiple CIs in the same repository, as well as a `testing` input for skipping the steps that require Vault or GCS access

Part of #296


academo (approved):
CI setup looks good to me. I didn't review the test files (assumed they were clean auto-generated)

PR #341, opened 2025-10-07
feat: add cd summary step

asimpson:
Convenient to get the full version if argo is skipped, e.g. want to deploy a one-off version to a single dev stack for testing.


xnyo (approved):
Great addition! 🙌  Tested [here](https://github.com/grafana/grafana-pluginsplatformprovisioned-app/actions/runs/18374937004) and it works properly. Only one small suggestion:

PR #350, opened 2025-10-14
fix(ci): allow-unsigned package creation on 'none' environment

theSuess:
The examples use the following logic to decide the environment:

```yaml
      environment: ${{ (github.event_name == 'push' && github.ref_name == 'main') && 'dev' || 'none' }}
```

This results in an `environment` value of `none` which no longer works after PR #300.

This PR fixes the issue by adding 'none' as a valid environment to skip signing for. Alternatively, we can update the examples. This would require updates in the consuming repos as well which is why I fixed it at this level.


xnyo (approved):
LGTM!

PR #356, opened 2025-10-15
feat: adding a non-required input called signature-type

tolzhabayev:
Adding a new input called `signature-type` which will be passed down to the `@grafana/sign-plugin` call. 
This input should be used when using this workflows but signing a plugin with a different signature type than `grafana` - for example:`community`

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -93,6 +93,11 @@ on:
         type: boolean
         required: false
         default: false
+      signature-type:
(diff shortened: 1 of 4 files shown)


xnyo (approved):
LGTM! 🚀 

Tested here (community): https://github.com/grafana/business-charts/actions/runs/18528477699

and here (grafana): https://github.com/grafana/plugins-drone-to-gha/actions/runs/18528487204

and the artifacts are signed with the correct signature level

PR #360, opened 2025-10-22
feat(ci): add new gha for running frontend e2e tests against specific stack

### LINKED ISSUE #306: CI: Run e2e tests in Grafana Cloud Dev
Add a way to run e2e tests on a Grafana Cloud Dev instance. Adapt and port this PR: https://github.com/grafana/shared-workflows/pull/1340 into plugin-ci-workflows instead, as it's better suited in this repository.

yduartep:
create new shared action to run e2e tests against pre-selected stack. You define in which region the selected stack belong, the plugin from where are executed the tests and optionally which other plugins and datasources you want to provision in that stack when starting a Grafana instance. You need to have the playwright configuration and the test specifications in the plugin that run the tests.

**Demo**:
https://github.com/user-attachments/assets/8fb3f278-a340-4a38-b588-1c685dab941b

Part of https://github.com/grafana/grafana-community-team/issues/632

Fixes #306


actions/internal/plugins/frontend-e2e-against-stack/action.yml
xnyo: The PR doesn't change the shared workflows, so I think the intent is having the plugin authors adding this action manually in their existing workflows, rather than running it as part of the shared workflows. If this is the case, the action should be moved in `actions/plugins/frontend-e2e-test-against-stack`, outside of the `internal` folder.

The `internal` folder is for actions used by the provided shared workflows. For actions that are to be used separately, the `actions/plugins` folder should be used instead.


If instead we should run this as part of the shared CI workflows (which could also make sense, but this PR doesn't make any changes to the workflows), they should live in `actions/internal`

actions/internal/plugins/frontend-e2e-against-stack/action.yml
```diff
@@ -0,0 +1,296 @@
...
+    - name: Checkout repository
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      with:
+        persist-credentials: false
+
+    - name: Setup Node.js environment
+      uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+      with:
+        node-version: "20"
+        cache: "yarn"
```
xnyo: These two may need to be configurable but we can keep them as-is for now until it's needed

actions/internal/plugins/frontend-e2e-against-stack/package.json
xnyo: Don't we have to commit the `yarn.lock` file as well?

xnyo (commented):
Awesome work! 🥳  Sorry for the lengthy review time on this. I don't have lots of knowledge on how plop works, but I see the current approach works fine for csp-app so I think we can merge it in plugin-ci-workflows mostly as-is.

I have left some comments, mostly around repo conventions, best practices, naming and other minor things. To speed things up since this has already been in review for a while, I have made a new branch and applied most of my suggestions here: https://github.com/grafana/plugin-ci-workflows/compare/yduartep/create-gha-for-e2e-tests-against-stack...giuseppe/e2e-cloud-pr-review

The review comments which have a corresponding fix commit in my branch have been marked accordingly. Please feel free to cherry-pick the commits from my branch into yours, or merge the whole branch into yours directly if you agree with my suggestions.

The things that haven't been done in my branch since they are not convention/best-practices-related and require further discussions are:

- Moving the action outside of the `internal` folder (I think we should go for it!)
- Changes to the release-please manifest file (same as above)
- Committing the yarn.lock file to the repo (not sure about it, but I think it makes sense)
- README changes about the deployment_tools changes if `npm-registry-auth` is `true` (not a blocker and I am checking the docs from platform-productivity to see if we can simplify the OIDC access)

Let me know if you have any questions about the review!

xnyo:
> Hi @xnyo thanks a lot for the review. I am happy with all your changes. DO you think you can merge your branch and close mine?

@yduartep yes I can rebase this PR, update it with my changes and merge. I will take care of that! 👍

xnyo (approved):

PR #365, opened 2025-11-03
feat: add warning message when using rolling releases channel (`@main`)

### LINKED ISSUE #353: Print warning when using unpinned workflows
Investigate a way to check if the current plugin repository is using `@main` instead of pinning to a version, and print a warning in the GHA summary if that's the case. This should push users to pinning the workflows instead of using the rolling release channel.

xnyo:
Encourages using tagged releases instead of the rolling releases channel. Fixes #353.

Example run using rolling releases (warning): https://github.com/grafana/plugins-drone-to-gha/actions/runs/19031442995

Example run using tagged releases (no warning): https://github.com/grafana/plugins-drone-to-gha/actions/runs/19031298151

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -303,7 +303,7 @@ env:
 jobs:
   common:
     name: Common checks
-    uses: grafana/plugin-ci-workflows/.github/workflows/common-checks.yml@giuseppe/warn-on-rolling-releases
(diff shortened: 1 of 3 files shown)


academo (commented):
I understand where this is going but wouldn't this get lost among the noise?  how likely is someone to read this message?

someone will have to go, see the actions, click the correct one and then click the common-checks

academo (approved):

Issue #354, opened 2025-10-14
Run plugin-validator as part of CI


xnyo:
Run plugin-validator as part of the pipeline. It should be opt-in first, since it was not part of the old Drone pipeline and we never used it for internal (Grafana Labs) plugins afaik, so some checks may need to be adjusted.

PR #366, opened 2025-11-03
feat: run plugin-validator as part of CI (opt-in)

xnyo:
Adds support for running plugin-validator as part of CI when input `run-plugin-validator: true` is passed to the workflow.

Part of #354.

Example runs can be found in the smoke tests for this PR, which have been updated to run the validator.

For example: https://github.com/grafana/plugin-ci-workflows/actions/runs/19139877974/job/54701694400?pr=366

(see the annotations at the top)

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -191,7 +191,7 @@ on:
         type: boolean
         required: false
         default: false
-      plugin-validator-config-file:
+      plugin-validator-config:
         description: |
           Content of the plugin validator configuration file (yaml) to use.
           If not provided, the action will look for the file specified in `plugin-validator-config-path` input instead.
@@ -422,7 +422,7 @@ jobs:
       golangci-lint-version: ${{ inputs.golangci-lint-version }}
 
       run-plugin-validator: ${{ inputs.run-plugin-validator }}
-      plugin-validator-config-file: ${{ inputs.plugin-validator-config-file }}
+      plugin-validator-config: ${{ inputs.plugin-validator-config }}
       plugin-validator-config-path: ${{ inputs.plugin-validator-config-path }}
 
       run-playwright: ${{ inputs.run-playwright }}
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -194,18 +194,20 @@ on:
       plugin-validator-config:
         description: |
           Content of the plugin validator configuration file (yaml) to use.
+          It has higher priority than `plugin-validator-config-path` input.
           If not provided, the action will look for the file specified in `plugin-validator-config-path` input instead.
+          If neither is provided, a default configuration will be used.
         type: string
         required: false
         default: ""
       plugin-validator-config-path:
         description: |
           Path to the plugin validator configuration file (yaml) to use.
-          Defaults to `.plugin-validator.yaml`.
-          If the file is not present, the action will use a default configuration instead.
+          It will be used only if `plugin-validator-config` input is not provided.
+          If not provided, a default configuration will be used.
         type: string
         required: false
-        default: .plugin-validator.yaml
+        default: ""
 
       # Feature toggle: Artifacts attestation for build provenance
       attestation:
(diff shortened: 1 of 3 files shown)


.github/workflows/cd.yml
```diff
@@ -185,7 +185,29 @@ on:
...
         required: false
 
-      # Artifacts attestation for build provenance
+      # Feature toggle: plugin-validator
+      run-plugin-validator:
+        description: Whether to run plugin-validator.
+        type: boolean
+        required: false
+        default: false
+      plugin-validator-config-file:
```
academo: this name is rather confusing, `plugin-validator-config-path` and `plugin-validator-config-file` can be confused as both accepting a path

what if we call them:

`plugin-validator-config` and the other one `plugin-validator-config-path`

.github/workflows/ci.yml
```diff
@@ -438,6 +460,54 @@ jobs:
...
+            echo "${PLUGIN_VALIDATOR_CONFIG_FILE}" > .plugin-validator.yaml
+            PLUGIN_VALIDATOR_CONFIG_PATH=.plugin-validator.yaml
+            cat ${PLUGIN_VALIDATOR_CONFIG_PATH}
+          elif [ -f "${PLUGIN_VALIDATOR_CONFIG_PATH}" ]; then
+            echo "Using plugin-validator configuration file at path: ${PLUGIN_VALIDATOR_CONFIG_PATH}"
+          fi
+
+          if [ ! -f "${PLUGIN_VALIDATOR_CONFIG_PATH}" ]; then
+            PLUGIN_VALIDATOR_CONFIG_PATH=".plugin-validator.yaml"
+            echo "⚠️ ${PLUGIN_VALIDATOR_CONFIG_PATH} configuration file is missing. Providing a default one as fallback."
```
academo: This should not really be a warning, I expect it'll be far more common to not pass your own config file.

.github/workflows/ci.yml
```diff
@@ -438,6 +460,54 @@ jobs:
...
+            cat <<EOF > "${PLUGIN_VALIDATOR_CONFIG_PATH}"
+          global:
+            enabled: true
+            severity: warning
```
academo: we should just let this be as is, each validator using its default value

.github/workflows/ci.yml
xnyo: 👍 , done!

.github/workflows/ci.yml
```diff
@@ -438,6 +460,54 @@ jobs:
...
+          EOF
+          fi
+
+          echo "Using configuration:"
+          cat "${PLUGIN_VALIDATOR_CONFIG_PATH}"
+
+          mkdir -p /tmp/empty
+
+          # Do not run clamav because it takes too long (and because it would scan node_modules as well)
+          docker run --name=plugin-validator --pull=always \
```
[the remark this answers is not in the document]
academo: you are correct. semgrep and gosect require system install. osv-scanner uses the go library so works via npx

.github/workflows/ci.yml
```diff
@@ -438,6 +462,62 @@ jobs:
...
+            PLUGIN_VALIDATOR_CONFIG_PATH=".plugin-validator.yaml"
+            echo "${PLUGIN_VALIDATOR_CONFIG}" > ${PLUGIN_VALIDATOR_CONFIG_PATH}
+          elif [ -n "${PLUGIN_VALIDATOR_CONFIG_PATH}" ]; then
+            # User-provided configuration file path
+            echo "Using plugin-validator configuration file at path: ${PLUGIN_VALIDATOR_CONFIG_PATH}"
+            if [ ! -f "${PLUGIN_VALIDATOR_CONFIG_PATH}" ]; then
+              echo "::error title=plugin-validator: missing config file::${PLUGIN_VALIDATOR_CONFIG_PATH} configuration file is missing."
+              exit 1
+            fi
+            if [[ -d "${PLUGIN_VALIDATOR_CONFIG_PATH}" ]]; then
```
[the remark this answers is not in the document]
xnyo: fixed!

academo (approved):

PR #368, opened 2025-11-03
fix: fail when plugin version exceeds 190 chars

s4kh:
Version field should be less than or equal to 190

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -366,7 +366,7 @@ jobs:
 
           # Plugin version should not exceed 190 characters including suffix
           if [ ${#pr_version} -gt 190 ]; then
-            echo "❌Error: Plugin version with suffix exceeds 190 character limit"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -363,6 +363,10 @@ jobs:
           package_json_path="$PLUGIN_DIRECTORY/package.json"
           version=$(jq -r .version "$package_json_path")
           pr_version="$version+${PLUGIN_VERSION_SUFFIX}"
+
+          # Version in the DB is at most 190 characters
+          pr_version="${pr_version:0:190}"
```
s4kh: @xnyo Would there be a case when user creates too long version and we overwrite it in the package json and drift happens between plugin.json and package.json? Highly unlikely?

academo: we should fail CI in this case. this is almost like a silent fail

xnyo (approved):
LGTM! 🙌

PR #389, opened 2025-11-20
test: run smoke tests with act

xnyo:
Part of #296 

Introduces a testing framework for plugin-ci-workflows written in Go and based on nektos/act.

Before, we used to run smoke tests inside GitHub Actions directly. This new approach has several advantages:

- Ability to run tests locally
- Ability to make proper assertions on workflow runs (e.g.: checking that artifacts are uploaded to GCS, checking that the plugin ZIPs are valid, checking that the plugin gets deployed, etc), rather than just checking if the workflow passes
- Ability to make tests for workflow failures: we can create tests for cases where the workflow should fail, and test that it actually fail (before, we could only write tests for successful workflow runs, otherwise the status checks would fail)
- Ability to mock re-usable actions. We can use this to mock Vault access, catalog publishing, argo workflow triggers, etc, without running the real steps

But also some disadvantages:

- The tests now run sequentially rather than in parallel, resulting in higher execution time. Running multiple act instances on the same job (VM) via `t.Parallel()` doesn't work properly:
  - Docker container names clash and we cannot run variations of the same workflows in parallel
  - The artifacts server is run on the same port, so even if multiple act instances start in parallel, they complain about the artifacts server port being in use
  - Even when specifying different docker container names and artifact port for each test case, there are issues due to multiple act docker containers sharing the same volume used for tests, which causes the workflow to fail because the files change while multiple workflows are running

I'll investigate solutions for those problems above in follow-up PRs.

This PR:

- Introduces this new testing framework
- Ports the old smoke tests to this new framework
- Removes the old smoke tests (TODO)

Follow-ups: #397


s4kh (approved):
Nice one! Probably the biggest con is execution time due to not being able to run the actions in parallel, right?

Issue #397, opened 2025-11-24
Improvements to act-based testing


xnyo:
#389 switches the smoke test system from using GitHub Actions directly to running act inside GitHub Actions. This allows us to write Go tests and run assertions on the workflows that are normally not possible or complicated to achieve via yaml files.

We can further improve act-based testing with some additional features:

- [x] Parsing of GHA commands, annotations, outputs, etc, in order to run assertions on them
- [x] Stronger assertions in the smoke tests (e.g.: CI output) rather than just asserting on success/failure 
- [x] Mock GCS for testing uploading to GCS (which is required for GCOM catalog publishing)
- [x] Mock GCOM for testing catalog publishing steps
- [x] Mock Vault access for testing accessing to Vault
- [x] Mock Argo workflow triggers for testing CD
- [x] Mock "Build and test" plugins in anything that doesn't test building or testing the plugin, to speed up act test runs ("Build and test" is the slowest job apart from Playwright)
- [x] Introduce caching mechanism for the nektos/act docker image and act tooling docker volume to speed up execution

xnyo:
Closing as completed

PR #438, opened 2025-12-09
chore: build and package plugin in its directory, better support for artifact prefix

xnyo:
Second part of splitting #401 into smaller PRs.

First part: #417

As part of #397 , this PR:

- Adds ids to all the important steps in ci, to allow mocking them (referencing them by id) in act tests
- Better use of the `plugin-directory` input for act tests, to allow running CI on the testdata plugins via act in parallel on the same GHA runner/VM:
  - Before, the testdata plugin would be built in `/dist` (where `/` is the repo root) and the dist artifacts would always be downloaded to `/tmp/dist-artifacts`, even if the plugin does not live in the root of the repo. This is fine if in the runner we have one repo with one plugin. If we have multiple plugins in the repo and ONE runner building multiple plugins in parallel, all the different plugins would use the same `/dist` and `/tmp/dist-artifacts` directories and files from different test runs/plugins would clash with each other
  - With this PR, the plugin is built and packaged in `/<plugin_directory>/dist` (in other words, next to its `package.json` file), and the artifacts are stored in `/tmp/<dist-artfacts-prefix>dist-artifacts`. This allows building multiple testdata plugins on the same GHA runner (VM). Each testdata plugin uses its own folders and multiple plugins don't clash with each other.
- Other smaller improvements to the ci and plugin actions to better work with act tests

### DIFF
--- a/actions/internal/plugins/package/package.sh
+++ b/actions/internal/plugins/package/package.sh
@@ -110,6 +110,8 @@ for file in $(find "$backend_folder" -type f -name "${exe_basename}_*"); do
     # Copy all files but the executables
     mkdir -p "$plugin_id"
     rsync -a --exclude "${exe_basename}*" "$dist/" "$plugin_id"
+    # TODO: this instead of rsync
+    # find "$dist" -type f ! -name "${exe_basename}*" -exec cp --parents {} "$plugin_id" \;
 
     # Copy only the current executable
     cp "$dist/$file" "$plugin_id/$backend_folder"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -403,6 +403,7 @@ jobs:
           INPUT_TESTING: ${{ inputs.testing }}
 
       - name: Setup
+        id: setup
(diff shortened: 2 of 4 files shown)


actions/internal/plugins/package/package.sh
```diff
@@ -110,6 +110,8 @@ for file in $(find "$backend_folder" -type f -name "${exe_basename}_*"); do
     # Copy all files but the executables
     mkdir -p "$plugin_id"
     rsync -a --exclude "${exe_basename}*" "$dist/" "$plugin_id"
+    # TODO: this instead of rsync
+    # find "$dist" -type f ! -name "${exe_basename}*" -exec cp --parents {} "$plugin_id" \;
```
xnyo: Related to act testing: the Docker image used for act doesn't include `rsync` unlike GitHub Actions. We currently install it via `apt` in the setup step, but it's an unnecesary step and a waste of time. We can use `find` instead of `rsync`, which is already installed. Just adding a comment for now, but I will add this later on as an improvement

s4kh (approved):

PR #442, opened 2025-12-10
test(act): mock steps and jobs, add artifacts interface, add more tests

xnyo:
Final part of splitting #401 into smaller PRs.

First part: #417 
Second part: #438 

As part of #397, this PR:

- Adds a system for intercepting GHA output commands, storing and asserting on them
- Adds a system for mocking jobs and step:
  - The test runner will take a workflow as a template, copy it and all the other workflows yaml files that are being called from it (called "children" workflow in the code), patch the workflow(s) by replacing/removing jobs or steps, based on the test case condfiguration, encode the workflow and all children workflows to temporary yaml files on disk and run act on them. This allows us to have test-specific yaml files with all the modifications that we need for that test (e.g.: for speed, we can replace the "Test and build" job with a mock one that puts the file in the correct place, rather than testing and compiling the plugin each time)
- Adds implementation for mocking GitHub artifacts upload/download, based on the afero fs library
- Adds a mocked step for replacing the "Test and build" step with one that copies the pre-built mocked dist artifacts in the correct folder
- Changes how the act test runner is set up to allow spinning up multiple of them at the same time, running in parallel
  - Give an unique uuid to each workflow, job and runner, to ensure Docker container names and volumes do not clash with each other (act uses the workflow/job name as part of the Docker container name and volume names, so they must be unique for multiple tests to run in parallel)
  - Spins up the act artifact server on a random free port, rather than using a fixed port, so multiple act instances an run in parallel
- Adds tests for the "package" step, asserting on the resulting GHA artifacts
- Add more assertions to the existing smoke tests, based on the resulting GHA artifacts
- Updates all tests to run in parallel
- Adds many utilities function for running assertions on tests, mostly around assserting for files in an afero.Fs (GHA artifacts)


tests/act/main_test.go
```diff
@@ -1,40 +1,21 @@
...
+	"strings"
 	"testing"
 
 	"github.com/grafana/plugin-ci-workflows/tests/act/internal/act"
 	"github.com/grafana/plugin-ci-workflows/tests/act/internal/workflow"
-	"github.com/stretchr/testify/require"
+	"github.com/spf13/afero"
 )
 
-func TestSmoke(t *testing.T) {
```
xnyo: Test case moved to `main_smoke_test.go`

tests/act/internal/workflow/plugins.go
xnyo: Moved to `ci.go` and updated to use the new children/mock system

s4kh (approved):

PR #464, opened 2025-12-18
test(act): add tests for workflow context evaluation

xnyo:
Adds act tests for workflow context evaluation.


s4kh (approved):
Nice! Functional options pattern 💯

PR #496, opened 2026-01-12
feat(ci): support react image in e2e matrix

sunker:
This PR updates the `e2e-version` action to v1.2.0. By doing so, the `e2e-versions` action will include a Grafana image built with react 19. This allows devs to test whether their plugins are forwards-compatiblity, as Grafana will move to React 19 in Grafana 13. 

The Grafana React image is added to the matrix by default, but users of plugin-ci-workflows will be able to disable it by setting the `skip-grafana-react-19-preview-image` to true.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -123,7 +123,7 @@ on:
         description: "Optionally, you can skip the Grafana React 19 preview image"
         type: boolean
         required: false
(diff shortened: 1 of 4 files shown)


tolzhabayev (approved):

xnyo (approved):
LGTM, awesome! 🙌

PR #498, opened 2026-01-12
test(act): refactor testing framework to handle more than one type of workflow

xnyo:
Part of #397.

Preliminary refactoring work for CD testing support. This refactoring extracts some implementation details from SimpleCI to the generic TestingWorkflow struct, which can then be embedded by the CD testing workflow implementation. CD testing implementation will come in a future PR.

Mostly moving stuff around, and changing the APIs a little bit to move what makes sense from the old `SimpleCI` testing workflow up to the base `TestingWorkflow` struct. **I added comments in the PR explaining the changes and why we need them. They seem to be collapsed by default in the new GitHub review UI, so please expand them while reviewing the PR, or click "Comments" here**

immagine


tests/act/internal/workflow/ci/mock.go
```diff
@@ -0,0 +1,27 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/grafana/plugin-ci-workflows/tests/act/internal/workflow"
+)
+
+// mockWorkflowContextStep returns a Step that mocks the "workflow-context" step
+// to return the given mocked Context.
+func mockWorkflowContextStep(ctx Context) (workflow.Step, error) {
```
xnyo: Same as before, just moved into the new `ci` package because this is ci-specific and not re-usable across all workflows.

tests/act/main_context_test.go
```diff
@@ -40,17 +41,19 @@ func TestContext(t *testing.T) {
...
+					Testing: workflow.Input(tc.testingInput),
+				}),
 				// Only run test-and-build job and stop after workflow-context step
 				// (no need to build the plugin, etc, for this test)
-				workflow.WithOnlyOneJob(t, testAndBuild),
-				workflow.WithRemoveAllStepsAfter(t, testAndBuild, workflowContext),
+				ci.MutateCIWorkflow().With(
+					workflow.WithOnlyOneJob(t, testAndBuild),
+					workflow.WithRemoveAllStepsAfter(t, testAndBuild, workflowContext),
+				),
```
xnyo: This is needed to re-use the "mutations" of the testing workflows (removing, skipping, no-opping and mocking/replacing steps). They used to be methods of `SimpleCI`, but we also need them for CD testing workflows as well, so they have been moved to the base `TestingWorkflow` struct, and we now provide this API for mutating workflows in test cases

s4kh (approved):

PR #206, opened 2025-08-05
feat(playwright): env for GRAFANA_VERSION to allow the playwright config access

L2D2Grafana:
Pass the `GRAFANA_VERSION: ${{ matrix.GRAFANA_IMAGE.VERSION }}` and `GRAFANA_IMAGE: ${{ matrix.GRAFANA_IMAGE.NAME }}` to playwright so that people can access it in the playwright config with `process.env.GRAFANA_VERSION` . The goal is to run different tests for different grafana versions. [example PR ](https://github.com/grafana/logs-drilldown/pull/1479). Right now `process.env.GRAFANA_VERSION` is undefined 
Screenshot 2025-08-05 at 9 55 11 AM.

PR #507, opened 2026-01-16
feat(playwright)!: remove version and image arguments

sunker:
Passing grafana version to Playwright has side effect, as plugin-e2e uses the version environment variable (if available) when resolving e2e-selectors. 

# ⚠️ **Breaking change**
For consumers of the shared workflows that wants to detect Grafana version within plugin-e2e tests, there's an alternative approach using the [official API](https://grafana.com/developers/plugin-tools/e2e-test-a-plugin/): 

```ts
test('should be possible to enable advanced mode', async ({ grafanaVersion, alertRuleEditPage }) => {
   test.skip(semver.lt(grafanaVersion, '11.6.0'), 'Advanced mode is not supported in Grafana versions < 11.6.0');
   // test code
});
 ```

This PR reverts the feature introduced in https://github.com/grafana/plugin-ci-workflows/pull/206


xnyo (approved):
LGTM!

PR #517, opened 2026-01-21
test(act): add tests for tooling versions (go, node) and js package manager detection logic

xnyo:
Adds some tests for the logic that handles Go/Node versions installation and js package manager detection in CI.

Also adds some utility testing workflow options to inject custom steps before/after pre-existing steps. This is different than replacing (mocking) a step, and can be used to manipulate the test environment via bash scripts before running some steps. (e.g.: it was used to delete .nvmrc before determining the Node version, to test its behavior)


tests/act/main_cache_warmup_test.go
xnyo: This shouldn't be in the diff (?)

s4kh (approved):

PR #538, opened 2026-01-26
fix(renovate): fix regexManagers for tooling versions

xnyo:
Should f.ix #537. Also replaces some deprecated options with supported ones.


sunker (approved):
🚀

PR #566, opened 2026-02-16
chore(ci): pin plugin-validator to 0.37.1 rather than using the latest version from main

### LINKED ISSUE #565: Pin plugin validator
We are currently always using the latest `main` with the plugin-validator.

Before running it by default (currently opt-in), we should switch to running stable releases instead.

We already cut releases in plugin-validator: https://github.com/grafana/plugin-validator/releases

Combined with renovate custom managers (e.g.: https://github.com/grafana/plugin-ci-workflows/blob/0e0617665e0b6acf4486f76121c51bebba77127a/renovate.json#L38) we can easily pin the validator and manage its dependency bumps via renovate.

This would avoid random failures like this one when new changes are merged into plugin-validator@main: https://github.com/grafana/plugin-ci-workflows/pull/564

xnyo:
Pin plugin-validator, add tests and renovate config.

Fixes #565


academo (approved):
code looks correct but I wonder if we can fix this better by crewating a `latest` tag in the validator. Do we risk people hardcoding validator versions and missing on new checks?

academo (approved):

PR #576, opened 2026-02-26
chore(deps)!: remove "v" prefix from mage-version input

xnyo:
Removes the "v" prefix from the mage version input for consistency with the other inputs that handle the tooling versions


mckn (approved):
LGTM!

PR #590, opened 2026-03-09
fix: remove gcs artifact exists step

s4kh:
Fixes https://github.com/grafana/plugin-ci-workflows/issues/583
Always upload.
The plugin version ZIP files sent to GCOM as URLs are copied from integration-artifacts to its own grafana-plugins-catalog[-dev|-ops] bucket by GCOM (https://github.com/grafana/grafana-com/issues/14289). This was not in place when plugin-ci-workflows was originally built. Due to this, I think we can safely ALWAYS overwrite files in integration-artifacts, so we can skip the exist check completely and always allow overwrite the files (provided the ZIP files are copied to its own, read-only bucket and those files in integration-artifacts are not used anywhere else by GCOM, rather than just for downloading and re-uploading to a different bucket).

With this proposed approach, attempting to publish a plugin version whose version value already exists in the catalog, the file in integration-artifacts file would be overwritten by plugin-ci-workflows, but GCOM will reject the plugin version with "plugin version already exists" error. Provided the integration-artifacts file is not used anywhere else, it should be safe.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -314,7 +314,7 @@ on:
           Branch or tag to publish from.
           Can be used to deploy PRs to dev.
           Can also be used to deploy specific branches (e.g.: release branches) or tags (e.g.: release tags) to any target environment.
-          In order to deploy to `prod` or `prod-canary`, where PRs or unreleased changes should not be deployed, 
+          In order to deploy to `prod` or `prod-canary`, where PRs or unreleased changes should not be deployed,
           the branch or tag must match the `release-reference-regex`, otherwise the workflow will fail.
           Defaults to `main`.
         default: main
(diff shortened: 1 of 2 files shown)


s4kh:
Act tests failing with 
plugin validator error:  **couldn't extract plugin archive: zip: not a valid zip file**
```
TestValidator/simple-frontend-yarn_succeeds_with_warnings: [CI/Plugins - CI/Test and build plugin    ] couldn't extract plugin archive: zip: not a valid zip file
TestValidator/simple-frontend-yarn_succeeds_with_warnings: [CI/Plugins - CI/Test and build plugin    ] ❌  Failure - Main plugin-validator [862.273862ms]
TestValidator/simple-frontend-yarn_succeeds_with_warnings: [CI/Plugins - CI/Test and build plugin    ] exitcode '1': failure
```

xnyo (commented):
Some comments, also commenting on the original issue with more findings

xnyo (approved):
LGTM!

PR #673, opened 2026-04-15
fix(test): allow hyphenated output keys to be mocked

oshirohugo:
## Summary

This allows `MockOutputsStep` to work with hyphenated output keys.

Before this fix, `MockOutputsStep` generated bash commands like:

```bash
echo "workflow-context=${workflow-context}" >> "$GITHUB_OUTPUT"
```

Bash does not support hyphens in variable names via `${...}` expansion. It silently parses `${workflow-context}` as `${workflow:-context}` (the "use default if unset" syntax) and since `$workflow` is always unset in the runner environment, the output becomes:

```
workflow-context=context
```

...instead of the intended value provided to `MockOutputsStep` through the map. No error, no warning, just a wrong value. Any caller passing a hyphenated key to `MockOutputsStep` (or `WithNoOpJobWithOutputs`, which calls it internally) got silently corrupted output.

This affects real-world actions with hyphenated output keys, most notably `actions/cache` (`cache-hit`), `actions/upload-artifact` (`artifact-id`, `artifact-url`), and the `workflow-context` step output used in CD gate checks.

## Changes

Replace hyphens with underscores in the env var name while keeping the original key in the `GITHUB_OUTPUT` line. Both the map key and the bash variable name are sanitized together:

```go
// Before
fmt.Sprintf(`echo "%s=${%s}" >> "$GITHUB_OUTPUT"`, k, k)
// env[k] = v

// After
sk := strings.ReplaceAll(k, "-", "_")
fmt.Sprintf(`echo "%s=${%s}" >> "$GITHUB_OUTPUT"`, k, sk)
// env[sk] = v
```

The original hyphenated key is preserved in the output name (left side), but the env var (right side) uses an underscore-safe name:

```bash
echo "workflow-context=${workflow_context}" >> "$GITHUB_OUTPUT"
```

## Tests

Added two test functions in `mock_test.go`:

**`TestMockOutputsStep`** — unit tests asserting the generated shell commands:
- Empty outputs map — still generates a valid no-op step
- Plain key without hyphens — regression guard, underscore form is unchanged
- Single-hyphen key (`workflow-context`) — the primary bug case; asserts `${workflow_context}` is used and `${workflow-context}` is absent
- Multiple-hyphen key (`test-and-build-output`) — broader coverage

**`TestMockOutputsStepGitHubOutput`** — integration tests that execute the generated bash script and assert on the actual `GITHUB_OUTPUT` file contents:
- Plain key, single-hyphen key, multiple-hyphen key, multiple mixed keys, and empty outputs


xnyo (commented):
Great catch! 🙌  However it looks like environment variables shouldn't have dashes at all, only letters, numbers and underscores:

> name A word consisting only of alphanumeric characters and underscores, and beginning with an alphabetic character or an underscore. Also referred to as an identifier.

I *think* the current implementation may work anyways because the env vars are passed by act itself, but maybe it's safer if we also replace "-" with "_" in `k` when setting the env for the step on line 77 (`env[k] = v` -> `env[toSnake(k)] = v`), wdyt?

xnyo (approved):
LGTM, CI failure is for unrelated changes. Failing due to a bug with the test case itself. It has to be fixed in a separate PR

xnyo:
Opened a PR here to fix failing CI (unrelated): https://github.com/grafana/plugin-ci-workflows/pull/678

PR #674, opened 2026-04-15
fix(ci): increase default artifact retention and allow config

### LINKED ISSUE #305: Extend retention period for dist-artifacts
dist-artifacts (built plugin) has a retention period of 7 days:

https://github.com/grafana/plugin-ci-workflows/blob/2d617a5adc492cae4668d1455d7ac8f2592de0c7/.github/workflows/ci.yml#L421

This artifact is used for GCS upload and, more importantly e2e tests.

This can cause issue if the workflow fails and the failing steps are re-run after 7 days (especially for the e2e tests, which are prone to failing).

For example: https://github.com/grafana/grafana-dbo11y-app/actions/runs/17735353829/job/51204238118

Image

The workaround is retrying the workflow from start to finish, which will force the plugin to be re-built and re-uploaded (so reset the retention period). However, the retention period should be extended to allow retrying e2e tests after 7 days from the original workflow run. It would also be nice to print a message explaining that the workflow should be re-run instead of this generic `Artifact not found for name` message, which is not clear.

Slack thread with more details: https://raintank-corp.slack.com/archives/C01C4K8DETW/p1758782856322209

oshirohugo:
## Summary

This PR increase the default retention period from 7 to 10 days and allow the users to input a custom artifact retation.
Fixes #305 

### Changes

- **Configurable retention period**: adds a `dist-artifacts-retention-days` input to `ci.yml` and `cd.yml` (default: 10 days). Teams that regularly retry jobs after long delays can raise this as needed.
- **Clearer error message**: the download step in `upload-to-gcs`, `playwright`, and `playwright-docker` now uses `continue-on-error: true` followed by an explicit failure step that emits a `::error::` annotation explaining that the artifact expired and that the workflow must be re-run from the beginning.

## Tests
- `TestDistArtifactsUnavailable` verifies both the default and a custom retention value, asserting that the workflow fails with the expected error annotation when the artifact is unavailable.
- A run of this branch can be seen here: https://github.com/grafana/plugins-drone-to-gha/actions/runs/24459902098/job/71470881932


.github/workflows/ci.yml
```diff
@@ -770,11 +778,22 @@ jobs:
...
           path: /tmp/${{ inputs.dist-artifacts-prefix }}dist-artifacts
 
+      - name: Log unavailable dist-artifacts error
+        if: steps.download-dist-artifacts.outcome == 'failure'
+        run: |
+          echo "::error::The dist-artifacts artifact could not be downloaded. It may have expired (retention period: ${RETENTION_DAYS} days). Please re-run the entire workflow from the beginning to rebuild the plugin."
+          exit 1
+        shell: bash
+        env:
+          RETENTION_DAYS: ${{ inputs.dist-artifacts-retention-days }}
```
xnyo, line 795: Nice! Can we add this check also to the other `download-artifact` calls in `cd.yml`?

https://github.com/grafana/plugin-ci-workflows/blob/59a0b015d37c52d8f98a7eacd40abd83801d1853/.github/workflows/cd.yml#L811-L815
https://github.com/grafana/plugin-ci-workflows/blob/59a0b015d37c52d8f98a7eacd40abd83801d1853/.github/workflows/cd.yml#L1046-L1050
https://github.com/grafana/plugin-ci-workflows/blob/59a0b015d37c52d8f98a7eacd40abd83801d1853/.github/workflows/cd.yml#L1259-L1263

xnyo (commented):
Great work! One small suggestion

xnyo (approved):
LGTM!

PR #676, opened 2026-04-16
fix: fail check-playwright-status on e2e tests timeout

oshirohugo:
## Summary

This PR fix the `check-matrix-status` to also consider github timeouts when running e2e and show a failure instead of pass when this timeout scenerion occurs in the Playwright e2s test.

Closes #567

## Changes

- **`actions/internal/check-matrix-status/action.yml`**: Extended the failure condition to also catch `cancelled` results, in addition to `failure`. The action now exits with a non-zero status and prints a descriptive message when either condition is met.

## Tests
- Added `tests/act/main_check_matrix_status_test.go` with a table-driven test covering all three result states: `success` (expects pass), `failure` (expects fail), and `cancelled` (expects fail).
- This PR was tested in a test plugin repository and the runs are here:
  - https://github.com/grafana/plugins-drone-to-gha/actions/runs/24664790296 (success)
  - https://github.com/grafana/plugins-drone-to-gha/actions/runs/24516498461 (timeout failure)


xnyo (approved):
LGTM!

PR #677, opened 2026-04-16
feat(ci): Use setup action in playwrite.yml

### LINKED ISSUE #447: frontend: reduce repetition of pnpm/action-setup
We should refactor playwright.yml to avoid repeated usage of pnpm/action-setup there and in the setup action.

Potential solution:

- calling the setup action in playwright.yml
- add an input (such as `frontend-only`) to the setup action to skip backend setup.

[See comment in PR 204](https://github.com/grafana/plugin-ci-workflows/pull/204#discussion_r2613722272)

Ukochka:
Replaces the duplicated package manager detection, pnpm installation, and Node.js setup steps in playwright.yml with a call to the shared setup action.

Fixes: #447


.github/workflows/playwright.yml
```diff
@@ -117,25 +117,14 @@ jobs:
...
-        if: steps.package-manager.outputs.name == 'pnpm'
-        uses: pnpm/action-setup@fc06bc1257f339d1d5d8b3a19a8cae5388b55320 # v5.0.0
-        with:
-          version: ${{ steps.package-manager.outputs.version }}
-
-      - name: Setup Node.js environment
-        uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
+      - name: Setup
+        id: setup
+        uses: grafana/plugin-ci-workflows/actions/internal/plugins/setup@main
```
leventebalogh, line 122: Just a heads up, we probably would like to pin this to a certain version with a SHA instead of pointing to main once merged, right?

actions/internal/plugins/setup/action.yml
```diff
@@ -42,12 +42,20 @@ inputs:
       Either node-version or node-version-file should be provided.
     required: false
     default: ""
+  frontend-only:
+    description: |
+      If true, skip backend (Go) setup including Go, mage, and golangci-lint installation.
```
leventebalogh: I would probably say `If "true" (...)` to make it clearer that it accepts a string. Or start with 
`Accepts the string "true" or "false" (default)."`.

leventebalogh (approved):
LGTM 👍

.github/workflows/playwright.yml
Ukochka: the release process (in release-please-pr-update-tagged-references.yml) automatically switches these to the tagged version when we cut a release.

PR #695, opened 2026-04-24
feat(playwright)!: bump e2e-version to v2.0.0

sunker:
Bumps `grafana/plugin-actions/e2e-version` from v1.2.1 to v2.0.0.

## ⚠ Breaking change

Plugins will now test against `grafana-enterprise:nightly` instead of the `grafana-dev` image. No workflow file changes are needed, but test results may differ.

## What changed

e2e-version v2.0.0 switches the nightly Grafana image from `grafana-dev` to `grafana-enterprise:nightly` and renames the controlling input from `skip-grafana-dev-image` to `skip-grafana-nightly-image`.

## Changes in this repo

- `playwright.yml`, `playwright-docker.yml`: bump action to v2.0.0, add `skip-grafana-nightly-image` input, pass `nightly-image || dev-image` to the action
- `ci.yml`, `cd.yml`: add `run-playwright-with-skip-grafana-nightly-image` input alongside the deprecated `run-playwright-with-skip-grafana-dev-image`

The old `skip-grafana-dev-image` / `run-playwright-with-skip-grafana-dev-image` inputs are kept and still honoured, so existing plugin workflows continue to work without changes.

## Testing

To test before merging, point a plugin's CI temporarily at this branch:

```yaml
uses: grafana/plugin-ci-workflows/.github/workflows/ci.yml@feat/bump-e2e-version-v2
```

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -130,8 +130,13 @@ on:
         description: "Optionally, use this input to pass a semver range of supported Grafana versions to test against. This is only used when version-resolver-type is plugin-grafana-dependency. If not provided, the action will try to read grafanaDependency from the plugin.json file."
         type: string
         required: false
+      run-playwright-with-skip-grafana-nightly-image:
+        description: "Optionally, you can skip the Grafana nightly image"
+        type: boolean
+        required: false
+        default: false
       run-playwright-with-skip-grafana-dev-image:
-        description: "Optionally, you can skip the Grafana dev image"
+        description: "Deprecated: use run-playwright-with-skip-grafana-nightly-image instead"
         type: boolean
         required: false
         default: false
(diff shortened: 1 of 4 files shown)


.github/workflows/cd.yml
```diff
@@ -1373,6 +1378,32 @@ jobs:
...
+                core.setOutput('skip', 'true')
+              } else {
+                core.setFailed(`Release ${tag} exists but has no assets and cannot be updated (immutable). Delete the release at ${release.data.html_url} and re-run.`)
+              }
+            } catch {
+              core.setOutput('skip', 'false')
+            }
+        env:
+          PLUGIN_VERSION: ${{ fromJSON(needs.ci.outputs.plugin).version }}
+
```
xnyo: What's the purpose of this step? It feels like this is a separate thing that should not be part of this PR, plus the "skip" output doesn't seem to be used anywhere?

xnyo: Related to this maybe? https://github.com/grafana/plugin-ci-workflows/issues/582

sunker: Oh yes @xnyo, this accidentally made its way into this pr. Removed now - sorry about this!

.github/workflows/playwright.yml
```diff
@@ -96,9 +102,9 @@ jobs:
 
       - name: Resolve Grafana E2E versions
         id: resolve-versions
-        uses: grafana/plugin-actions/e2e-version@e2e-version/v1.2.1
+        uses: grafana/plugin-actions/e2e-version@73cd57e848b81e75e3c17697e6701c9fa1404bcc # e2e-version/v2.0.0
```
sunker, line 105: we should be using git cmt sha, right?

xnyo: In this case it doesn't matter: actions under the `grafana` org are considered trusted and can be pinned to anything and won't be flagged by zizmor. We usually pin everything to SHA so I'd keep like this for consistency 👍

xnyo (approved):
LGTM!

xnyo:
Act tests failure look like a transient error. I triggered a new run and it should succeed

PR #704, opened 2026-04-28
fix(cd): pass provenance attestation to catalog publish

### LINKED ISSUE #102: Provenance attestation not passed to publishing step
### Summary

The provenance attestation is being generated in the CI workflow ([see lines 320-325](https://github.com/grafana/plugin-ci-workflows/blob/72e5e7da6af37b379bfede7e87773922042b63a9/.github/workflows/cd.yml#L320-L325)), but is not currently passed on or used in the publishing call ([see publish.sh](https://github.com/grafana/plugin-ci-workflows/blob/main/actions/plugins/publish/publish/publish.sh)).

### Details
- **Attestation Generation:**
  The workflow creates a provenance attestation artifact, but the artifact is not propagated to the publish step or included in the plugin publishing process.
- **Missing Usage:**
  The publishing script (`publish.sh`) does not currently accept or use the attestation file, so the produced artifact is not uploaded or referenced during publishing.
- **Reference Implementation:**
  See [plugin-uploader reference](https://github.com/grafana/plugin-uploader/blob/674e03b3266e58fc3182134ffb29d5cf6948a690/pkg/gcomservice/gcomservice.go#L152C8-L152C29) for how attestation can be included in the publish process.

### Suggested Action
Update the publishing step to accept and use the provenance attestation, following the example in the plugin-uploader reference. This will ensure the provenance data is published alongside the plugin artifact.

tolzhabayev:
## Summary

- Generate a GCOM-compatible provenance attestation reference from the attested universal plugin ZIP.
- Make catalog publishing wait for the attestation job when attestations are enabled.
- Pass the attestation through the publish action and include it as `provenanceAttestation` in the GCOM publish payload.
- Add a focused act test that enables attestation and asserts the mocked GCOM publish request includes `provenanceAttestation`.

## Root Cause

The CD workflow generated a build provenance attestation, but the publishing path did not depend on that job and the publish action had no input or request-field support for forwarding the attestation to GCOM.

Fixes #102

## Validation

- `make actionlint`
- `bash -n actions/plugins/publish/publish/publish.sh`
- YAML parse check for `.github/workflows/cd.yml` and `actions/plugins/publish/publish/action.yml`
- Publish script dry-runs with and without `--provenance-attestation`
- Shell check of the universal-ZIP attestation reference generation block
- `git diff --check`
- `GOCACHE=/tmp/plugin-ci-go-cache GOMODCACHE=/tmp/plugin-ci-gomodcache GOTOOLCHAIN=auto go test -short -count=1 -run "^$" ./...` from `tests/act`

Note: `SKIP_ACT_CACHE_WARMUP=1 GOTOOLCHAIN=auto go test -run TestCD_PublishIncludesProvenanceAttestation -count=1 -v` could not complete locally because Docker is not running (`Cannot connect to the Docker daemon at unix:///var/run/docker.sock`).

Also note: `make act-lint` could not run locally because the installed `golangci-lint` binary was built with Go 1.24 while `tests/act/go.mod` targets Go 1.25.4.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1004,7 +1004,14 @@ jobs:
     needs:
       - ci
       - publish-to-catalog
-    if: ${{ inputs.grafana-cloud-deployment-type != '' && inputs.trigger-argo }}
+    if: >-
+      ${{
+        !cancelled()
+        && needs.ci.result == 'success'
+        && needs.publish-to-catalog.result == 'success'
+        && inputs.grafana-cloud-deployment-type != ''
+        && inputs.trigger-argo
+      }}
 
     steps:
       - name: Check deployment type
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -806,6 +806,7 @@ jobs:
       attestation-id: ${{ steps.attestation.outputs.attestation-id }}
       attestation-url: ${{ steps.attestation.outputs.attestation-url }}
       bundle-path: ${{ steps.attestation.outputs.bundle-path }}
+      provenance-attestation: ${{ steps.provenance-attestation.outputs.provenance-attestation }}
 
     steps:
       - name: Download GitHub artifact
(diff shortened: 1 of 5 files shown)


xnyo (approved):
LGTM!

PR #705, opened 2026-04-28
fix(ci): gate GCS upload on Playwright success

### LINKED ISSUE #56: CI: Skip uploading to GCS if playwright tests fail
Currently, `.github/workflows/ci.yml`'s `upload-to-gcs` step is skipped if `jobs.test-and-build` fails, but is not skipped if `jobs.playwright` fails. In general, I think it'd be helpful to skip `upload-to-gcs` if `inputs.run-playwright` is `true` and `jobs.playwright` fails.

tolzhabayev:
## Summary

- Make `upload-to-gcs` wait for both Playwright jobs.
- Keep Playwright gating conditional so skipped jobs do not block GCS upload when the corresponding input is disabled.
- Add a workflow consistency test for the `upload-to-gcs` dependency and condition wiring.

## Root Cause

`upload-to-gcs` only depended on `test-and-build`, so it could still run after a Playwright job failed as long as the build job succeeded and the context was trusted.

Fixes #56

## Validation

- `make actionlint`
- `git diff --check`
- `GOCACHE=/tmp/plugin-ci-go-cache GOMODCACHE=/tmp/plugin-ci-gomodcache GOTOOLCHAIN=auto go test -short -count=1 -run TestCIUploadToGCSWaitsForEnabledPlaywrightJobs ./...` from `tests/act`
- `GOCACHE=/tmp/plugin-ci-go-cache GOMODCACHE=/tmp/plugin-ci-gomodcache GOTOOLCHAIN=auto go test -short -count=1 -run "^$" ./...` from `tests/act`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -763,11 +763,21 @@ jobs:
     name: Upload to GCS
     runs-on: ubuntu-arm64-small
 
-    # Skip the GCS upload (no access to the bucket) for PRs when run in non-trusted context (forks)
-    if: ${{ fromJson(needs.test-and-build.outputs.workflow-context).isTrusted }}
+    # Skip the GCS upload (no access to the bucket) for PRs when run in non-trusted context (forks).
+    # If Playwright is enabled, only upload after the corresponding Playwright job succeeds.
+    if: >-
+      ${{
+        !cancelled()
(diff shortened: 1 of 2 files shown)


toddtreece (approved):

sunker (approved):

xnyo (approved):
LGTM!

Issue #554, opened 2026-02-10
Add support for yarn 4


xnyo:
https://github.com/grafana/logs-drilldown/actions/runs/21841792325/job/63027761827?pr=1742

https://raintank-corp.slack.com/archives/C01C4K8DETW/p1770743718380439

```
Run grafana/plugin-ci-workflows/actions/internal/plugins/frontend@ci-cd-workflows/v4.0.0
  with:
    plugin-directory: .
    npm-registry-auth: false
  env:
    DEFAULT_NODE_VERSION: 22
    DEFAULT_GO_VERSION: 1.24
    DEFAULT_GOLANGCI_LINT_VERSION: 1.64.8
    DEFAULT_TRUFFLEHOG_VERSION: 3.90.6
    GCS_ARTIFACTS_BUCKET: integration-artifacts
    VAULT_INSTANCE: ops
    GOTOOLCHAIN: local
Run /opt/actions-runner/_work/_actions/grafana/plugin-ci-workflows/ci-cd-workflows/v4.0.0/actions/internal/plugins/frontend/pm.sh install
  /opt/actions-runner/_work/_actions/grafana/plugin-ci-workflows/ci-cd-workflows/v4.0.0/actions/internal/plugins/frontend/pm.sh install
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
  env:
    DEFAULT_NODE_VERSION: 22
    DEFAULT_GO_VERSION: 1.24
    DEFAULT_GOLANGCI_LINT_VERSION: 1.64.8
    DEFAULT_TRUFFLEHOG_VERSION: 3.90.6
    GCS_ARTIFACTS_BUCKET: integration-artifacts
    VAULT_INSTANCE: ops
    GOTOOLCHAIN: local
    PACKAGE_MANAGER: 
Running 'install' with yarn...
error This project's package.json defines "packageManager": "yarn@4.5.0". However the current global version of Yarn is 1.22.22.

Presence of the "packageManager" field indicates that the project is meant to be used with Corepack, a tool included by default with all official Node.js distributions starting from 16.9 and 14.19.
Corepack must currently be enabled by running corepack enable in your terminal. For more information, check out https://yarnpkg.com/corepack.
Error: Process completed with exit code 1.
```

xnyo:
[...]

PR #746, opened 2026-05-22
chore(security): test plugins: harden security and update packages

xnyo:
Hardens the test plugins used within act tests with the latest security practices:

- ignore post-install scripts, disallow git, minReleaseAge = 3 across npm, pnpn and yarn (v1, not all options are supported. yarn berry/v4 migration pending) test setups
- pin packages to exact versions in package.json
- pin grafana packages to nightly to avoid transitive git dependency
- update Go to latest version to address some Go stdlib CVEs in simple-backend test plugin
- `npx @grafana/create-plugin update` across all test plugins
- fix high and critical vulnerabilities highlighted by osv-scanner across all test plugins by bumping/overriding packages
- re-generates all mockdata for act tests using the latest artifacts built after all the changes above

Fixes https://github.com/grafana/grafana-catalog-team/issues/926

Follow-up: Upgrade the yarn test setup from v1 to v4 which supports all the supply-chain attack policies: #554


.github/workflows/ci.yml
```diff
@@ -348,7 +348,7 @@ permissions:
 env:
   # Default versions for tooling
   DEFAULT_NODE_VERSION: "24"
-  DEFAULT_GO_VERSION: "1.26"
+  DEFAULT_GO_VERSION: "1.26.3"
```
xnyo, line 351: SDK requires this patch specifically: https://github.com/grafana/grafana-plugin-sdk-go/blob/ff11f4c31e84754c6351e05736eb7cb8160dce38/go.mod#L3

If we losen the dependency to just `1.26`, tests fail due to some Act weirdness around caching.

Note: this is used as a very last resort. The go version in `go.mod` has higher priority, and that's pretty much a requirement since Go 1.11

scripts/mockdata-dist.sh
```diff
@@ -33,6 +33,10 @@ echo "Using Package Manager: $(pm)"
 echo "[$1] (frontend) Installing"
 $(pm) install
 
+echo "[$1] (frontend) Lint + typecheck"
+$(pm) run lint
+$(pm) run typecheck
```
xnyo, line 38: These are done in CI anyways. Added as an extra sanity check before building and packaging the mockdata files

andresmgot (approved):

PR #747, opened 2026-05-25
chore(deps): bump actions/upload-artifact to v7.0.1 and actions/download-artifact to v8.0.1

xnyo:
Bumps actions/upload-artifact to v7.0.1 and actions/download-artifact to v8.0.1. These versions support Node 24 instead of Node 20, which will be deprecated on 20260602. No other breaking changes.


xnyo:
https://github.com/nektos/act/pull/6039

PR #772, opened 2026-06-04
feat(playwright): add playwright-browsers input for E2E installs

CatPerry:
## Summary

Adds a `playwright-browsers` input (default `chromium`) to `cd.yml`, `ci.yml`, and `playwright.yml`. Plugins list exactly which browsers to install before Playwright E2E tests — space-, newline-, or semicolon-separated (e.g. `chromium firefox`).

The upper-level workflows (`cd.yml`, `ci.yml`) only pass through that one string. Browser-specific install logic (allowlist validation, per-browser install) lives entirely in `playwright.yml`.

## Problem

The shared workflows only install **Chromium** via `playwright install --with-deps chromium`. This can silently work for plugins that also run Firefox tests — self-hosted runners sometimes have an old Firefox binary cached from previous jobs.

When **Renovate bumps `@playwright/test`** (e.g. 1.59 → 1.60), Playwright expects **new browser builds** under new cache paths (like `firefox-1522` for 1.60). CI downloads Chromium for the new version, but never downloads Firefox. The old cached Firefox from 1.59 doesn't match anymore. Every Firefox test fails with:

```
browserType.launch: Executable doesn't exist at /home/ubuntu/.cache/ms-playwright/firefox-1522/firefox/firefox
```

Chrome tests pass fine. This is what happened on [grafana-collector-app#1698](https://github.com/grafana/grafana-collector-app/pull/1698) — 51 Firefox failures, 68 Chromium passes.

## Design

| Layer | What it knows |
|-------|---------------|
| `cd.yml` | `playwright-browsers: string` — pass-through only |
| `ci.yml` | Same |
| `playwright.yml` | Parses input, validates tokens, installs each listed browser |

**Install behavior in `playwright.yml`:**

| Browser | When listed | On install failure |
|---------|-------------|-------------------|
| `chromium` | Default; always if list is empty | Job fails |
| `firefox` | When listed | Job fails |
| `webkit` | When listed | **Best-effort** — system libs (GTK4, GStreamer) may be missing on self-hosted runners |

Which browsers actually **run** is still controlled by each plugin's `playwright.config.ts` (e.g. collector-app skips WebKit when `CI` is set).

## Why an allowlisted string, not raw interpolation

An earlier draft interpolated a free-form string directly into `run:`, which is a script-injection risk. The final design passes the input via `PLAYWRIGHT_BROWSERS` env and validates each token against `chromium`, `firefox`, `webkit` before running hardcoded `playwright install` commands.

Per [review feedback](https://github.com/grafana/plugin-ci-workflows/pull/772#pullrequestreview-4436700206), an explicit browser list is preferred over a single on/off flag so plugins install only what they need.

## Consumer change (after this is tagged)

```yaml
# In the plugin's ci-cd.yml
uses: grafana/plugin-ci-workflows/.github/workflows/cd.yml@ci-cd-workflows/vX.Y.Z
with:
  playwright-browsers: chromium firefox
```

Plugins pin shared CI to a release tag. Merging this doesn't help consumers until a new tag is cut.

## Test plan

- [x] Input wiring: `cd.yml` → `ci.yml` → `playwright.yml` passes `playwright-browsers` through.
- [x] Default (`chromium`): only Chromium installs — existing behavior for all consumers.
- [x] `playwright-browsers: chromium firefox`: Firefox install runs; Firefox failure fails the job.
- [x] `playwright-browsers: chromium webkit`: WebKit install is best-effort; job continues on WebKit failure.
- [x] Invalid token (e.g. `chrome`): job fails with allowlist error.
- [x] Cache key includes Playwright version and sorted browser list (e.g. `playwright-1.60.0-chromium-firefox`).
- [ ] After tag: verify on grafana-collector-app that E2E Firefox matrix passes with `playwright-browsers: chromium firefox`.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -105,13 +105,22 @@ on:
         type: boolean
         required: false
         default: false
-      playwright-browsers:
+      playwright-multi-browser:
         description: |
-          Space-separated browsers passed to `playwright install --with-deps`.
-          Match the browser projects in the plugin's playwright.config.ts (e.g. `chromium firefox`).
-        type: string
+          When true, install additional browsers (Firefox and WebKit) alongside Chromium.
+          Enable when the plugin runs non-Chromium browser projects on CI.
+          Which browsers actually run is controlled by the plugin's playwright.config.ts.
+        type: boolean
         required: false
-        default: chromium
+        default: false
+      playwright-multi-browser-blocking:
+        description: |
+          When false, failures in the additional browser install steps (Firefox, WebKit)
+          are non-blocking — the job continues and the plugin config skips browsers that
+          didn't install. Set to true if additional browsers must succeed or fail the build.
+        type: boolean
+        required: false
+        default: false
 permissions:
   contents: read
   id-token: write
@@ -212,8 +221,20 @@ jobs:
           path: ~/.cache/ms-playwright
           key: playwright-${{ steps.version.outputs.version }}
 
-      - name: Install Playwright Browsers
-        run: ${{ steps.setup.outputs.node-package-manager-exec-local-cmd }} playwright install --with-deps ${{ inputs.playwright-browsers }}
+      - name: Install Playwright Chromium
+        run: ${{ steps.setup.outputs.node-package-manager-exec-local-cmd }} playwright install --with-deps chromium
+        shell: bash
+
+      - name: Install Playwright Firefox
+        if: inputs.playwright-multi-browser
+        continue-on-error: ${{ inputs.playwright-multi-browser-blocking == false }}
+        run: ${{ steps.setup.outputs.node-package-manager-exec-local-cmd }} playwright install --with-deps firefox
+        shell: bash
+
+      - name: Install Playwright WebKit
+        if: inputs.playwright-multi-browser
+        continue-on-error: ${{ inputs.playwright-multi-browser-blocking == false }}
+        run: ${{ steps.setup.outputs.node-package-manager-exec-local-cmd }} playwright install --with-deps webkit
         shell: bash
 
       - name: Download GitHub artifact
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -105,6 +105,13 @@ on:
         type: boolean
         required: false
         default: false
+      playwright-browsers:
+        description: |
+          Space-separated browsers passed to `playwright install --with-deps`.
+          Match the browser projects in the plugin's playwright.config.ts (e.g. `chromium firefox`).
+        type: string
+        required: false
+        default: chromium
 permissions:
   contents: read
   id-token: write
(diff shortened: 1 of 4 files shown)


sunker (commented):
Makes sense to support more browsers than just chromium, but I think we should be explicit about which browsers rather than a single on/off flag. Maybe we can add a newline-separated or semicolon-separated `playwright-browsers` input that lists exactly what to install, defaulting to chromium so current behaviour is unchanged?  

We can still keep it injection-safe by passing the input via env and validating each token against an allowlist. This only controls what's installed - which browsers run stays up to the plugin's playwright.config.ts.

CatPerry:
> Makes sense to support more browsers than just chromium, but I think we should be explicit about which browsers rather than a single on/off flag. Maybe we can add a newline-separated or semicolon-separated `playwright-browsers` input that lists exactly what to install, defaulting to chromium so current behaviour is unchanged?
> 
> We can still keep it injection-safe by passing the input via env and validating each token against an allowlist. This only controls what's installed - which browsers run stays up to the plugin's playwright.config.ts.

I updated the changes to implement this. Thanks

.github/workflows/playwright.yml
```diff
@@ -205,9 +215,39 @@ jobs:
           path: ~/.cache/ms-playwright
           key: playwright-${{ steps.version.outputs.version }}
```
sunker: I guess you somehow need to include the browsers in the cache key

sunker (commented):
Added a comment. Think you also need to sign the CLA @CatPerry.

.github/workflows/playwright.yml
CatPerry: Fixed — added a `Resolve Playwright browsers` step that validates/dedupes the input and builds a sorted cache key (e.g. `chromium-firefox`). The cache key is now `playwright-<version>-<cache-key>`.

.github/workflows/playwright.yml
```diff
@@ -198,16 +208,59 @@ jobs:
...
 
-      - name: Install Playwright Browsers
-        run: ${{ steps.setup.outputs.node-package-manager-exec-local-cmd }} playwright install --with-deps chromium
+      - name: Install Playwright browsers
         shell: bash
+        env:
+          NPM_EXEC_CMD: ${{ steps.setup.outputs.node-package-manager-exec-local-cmd }}
+        run: |
+          set -euo pipefail
+          for token in ${{ steps.browsers.outputs.list }}; do
```
sunker: nit: nit: could we pull this through env: like the NPM_EXEC_CMD right above and the resolve step? something like
```yml
env:
  PLAYWRIGHT_BROWSERS_LIST: ${{ steps.browsers.outputs.list }}
  NPM_EXEC_CMD: ${{ steps.setup.outputs.node-package-manager-exec-local-cmd }}
run: |
  set -euo pipefail
  for token in $PLAYWRIGHT_BROWSERS_LIST; do
```

sunker (commented):
Just one nit, otherwise looks great!

.github/workflows/playwright.yml
CatPerry: 100%

CatPerry: Fixed — install step now uses `PLAYWRIGHT_BROWSERS_LIST` via `env:` and iterates with `$PLAYWRIGHT_BROWSERS_LIST`.

sunker (approved):
Thanks, great stuff @CatPerry!

PR #780, opened 2026-06-12
chore(version-bump-changelog): add pr creation and signed commits

L2D2Grafana:
## Summary 📝

- Replace `git commit` / `git push` and `git tag` / `git push --tags` in `version-bump-changelog` with the GitHub Git Data API so version bumps land on `main` as **verified GitHub App commits** and tags are created via the API.
- Allow committing `package.json`, `package-lock.json`, `yarn.lock`, `pnpm-lock.yaml`, and `CHANGELOG.md` in the API commit step.
- Document in `examples/extra/version-bump-changelog.yml` that this direct-to-main workflow is only for **private repos without a PR-required ruleset on main**.
- I'm not sure how to safely test workflow changes anymore with the new security rules https://github.com/grafana/deployment_tools/pull/617709 
- I'm also open to removing `version-bump-changelog` and updating the docs to migrate to release-please.


actions/plugins/version-bump-changelog/action.yml
```diff
@@ -92,24 +88,131 @@ runs:
...
+            'CHANGELOG.md',
+          ]);
+
+          const changed = execSync('git diff --name-only', { cwd: ws, encoding: 'utf8' })
+            .trim()
+            .split('\n')
+            .filter((file) => file && allowedFiles.has(file));
+
+          if (changed.length === 0) {
+            core.setFailed('No version bump files to commit');
```
xnyo: I think it should not fail to match the behaviour of the previous flow:

```
git add CHANGELOG.md || true  # No-op if changelog not generated
```

wdyt?

xnyo (commented):
Great work! I didn't test the changes but the code looks reasonable to me. I only have one suggestion

actions/plugins/version-bump-changelog/action.yml
L2D2Grafana: Updated! Let me know if there is a security sanctioned test repo for trying out new ci jobs these days and I'm happy to test it. Pretty sure it'll work though.

xnyo (approved):
LGTM! Great work! 🙌

PR #781, opened 2026-06-12
fix(cd): update softprops/action-gh-release to v3

L2D2Grafana:
## Description ✨
Hello it's me again 🐶 . I was so hoping that release-please would automate uploading to github releases and [it does](https://github.com/grafana/logs-drilldown/releases/tag/v2.1.0), BUT it doesn't create named/versioned zip files 😢.
Screenshot 2026-06-12 at 1 21 31 PM
Many of our open source users rely on [this zip file](https://github.com/grafana/logs-drilldown/issues/1784). My current plan is to use release-please but not have it update github-release. Then use the cd.yml with 
```
# Publish the github release
github-draft-release: false
``` 
Do let me know if there is a better way!

## Summary 📝

- Fixes [#582](https://github.com/grafana/plugin-ci-workflows/issues/582): bump `softprops/action-gh-release` from v2.6.2 to [v3.0.0](https://github.com/softprops/action-gh-release/releases/tag/v3.0.0) in `cd.yml` so GitHub releases on immutable-release repos upload assets before publish (handled upstream since v2.5; v3 aligns with the Node 24 Actions runtime).
- No CD workflow logic changes — `draft: ${{ inputs.github-draft-release }}` is unchanged; the draft-then-publish workaround was dropped in favour of relying on softprops' built-in behaviour.
- For prereleases on immutable repos, consumers should keep `github-draft-release: true` per [softprops/action-gh-release#763](https://github.com/softprops/action-gh-release/pull/763).

> [!NOTE]
> v3.0.0 requires Actions Runner ≥ 2.327.1 (Node 24 runtime). Grafana self-hosted runners (`ubuntu-arm64-small`, etc.) should already support this; if a release job fails with `using: node24 is not supported`, the runner fleet needs upgrading.

## Test 🧪

- [x] `make actionlint` passes

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -332,12 +332,7 @@ on:
         type: boolean
       github-draft-release:
         description: |
-          Keep the GitHub release as a draft after artifact upload. If disable-github-release is true, this will be ignored.
-
-          Defaults to true (draft). Set to false to publish the release after artifacts are uploaded.
-          On public Grafana repos with immutable releases, artifacts cannot be added to an already-published
-          release — CD always uploads to a draft first, then publishes when this input is false.
-          Do not publish before upload (e.g. via release-please without draft: true) or asset upload will fail.
+          Publish a draft release on GitHub. If disable-github-release is true, this will be ignored.
         default: true
         type: boolean
       disable-github-release:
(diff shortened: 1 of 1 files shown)


mckn (approved):
LGTM!

xnyo (approved):

PR #784, opened 2026-06-16
fix(ci): restrict GITHUB_TOKEN permissions in pr-checks-workflow-references

xnyo:
## Summary

- Add an explicit `permissions` block to the `check-references` job in `pr-checks-workflow-references.yml`
- Scope `GITHUB_TOKEN` to `contents: read` and `pull-requests: read` — the minimum required for checkout and the pull_request trigger context
- Fixes a `zizmor` `excessive-permissions` audit finding flagged in CI

Without this block the job inherits the org's default token permissions, which may be broader than needed. Explicit permissions make the scope clear and prevent the job from running with unintended write access if org defaults are ever widened.

## Test plan

- [ ] `Check @main workflow references` workflow still passes on a non-release-please PR
- [ ] `make actionlint` passes locally


mckn (approved):
LGTM!

PR #819, opened 2026-06-25
feat(playwright): migrate report publishing from plugin-actions to GCS

mckn:
## Summary

This pull request moves Playwright report publishing from `grafana/plugin-actions#253` into this repository. Two new actions live under `actions/plugins/playwright-reports/`:

- `upload-report-artifacts` runs once for each matrix entry. It renames the report directory after the Grafana image and version. It writes a `summary.txt` file. It then uploads the result as a GitHub artifact with one day of retention.
- `publish-report` runs once for each workflow run. It downloads all of those artifacts. It uploads them to Google Cloud Storage. It posts a pull request comment with a result table and links to the reports.

`playwright.yml` calls both actions. `ci.yml` and `cd.yml` gain one new input, `playwright-pr-comment-summary`.

## Why Google Cloud Storage replaces gh-pages

- The old action pushed reports to a `gh-pages` branch. That push needed `contents: write`. The new jobs need `id-token: write` and `pull-requests: write` instead.
- The old push created an unsigned commit. `grafana/plugin-tools` disabled its publish job for that reason. This change unblocks that job.
- Only Grafana employees can read the reports. Google serves them from `storage.cloud.google.com`.
- A 90-day lifecycle rule on the bucket deletes old reports.

## Storage and access

Reports go to `gs://grafana-e2e-test-artifacts/<owner>/<repository>/<date>/<pull-request-or-run-id>/`. The bucket lives in the `grafanalabs-global` project.

`publish-report` impersonates `github-e2e-test-artifacts@grafanalabs-workload-identity.iam.gserviceaccount.com`. That service account holds `roles/storage.objectUser` on this one bucket.

The service account trusts the `job_workflow_ref` attribute of this repository's Playwright workflows, pinned to `main` and to release tags. This is the same trust pattern that `github-plugin-ci-workflows@` already uses. The pattern gives three properties:

- Every repository that calls `playwright.yml` inherits access. No `deployment_tools` onboarding step exists.
- Every event works, including `push` and `schedule`. The attribute does not constrain the event.
- A workflow that a plugin repository writes itself cannot get the credential. Only this repository's pinned workflows can.

### Per-repository isolation

Please read this part carefully during review. The service account can write anywhere in the bucket. Each repository stays inside its own prefix because `publish-report` builds the destination from `github.repository`. A caller cannot set that value. IAM does not enforce the boundary, so four things must stay true:

- No input that a caller controls may reach the upload destination.
- The publish job must not check out the caller's repository.
- No step may execute content from the downloaded artifacts. Reading the files as data is safe.
- The upload must stay in its own job, separate from the job that runs the caller's tests.

`action.yml` and the action README both record these rules next to the login step.

## Changes that affect plugin repositories

- Fork pull requests do not publish reports. GitHub withholds the `id-token: write` permission from them.
- `publish-report` has no `bucket` input and no `service-account` input. Both values are internal to the action.
- `ci.yml` and `cd.yml` now request `pull-requests: write`. A called reusable workflow cannot exceed the permissions of its caller. Without this change the comment step fails with a 403 error.
- Report artifacts now expire after one day instead of 30 days. They exist only as input to the publish job. The reports themselves live in the bucket for 90 days.

## Release configuration

`release-please-config.json` registers both actions as separate packages. Each action then gets its own version tag. External consumers such as `grafana/plugin-tools` can pin a version.

## Tests

- `make actionlint` reports no errors.
- `make genreadme` reports no change.

These manual checks are still open:

- [ ] Run CI on a pull request and confirm that the comment shows working `storage.cloud.google.com` links.
- [ ] Confirm that objects appear under `gs://grafana-e2e-test-artifacts/<owner>/<repository>/<date>/<number>/`.
- [ ] Run CI on a push to a default branch and confirm that the publish job succeeds and writes the results table to the job summary.
- [ ] Confirm that a fork pull request skips the publish job.

## Dependencies and follow-up work

- `deployment_tools#712713` must create the bucket. Merge it before this pull request. It also needs rework: grant the service account plain `objectUser`, and move its impersonation members from the per-repository allowlist to `attribute.job_workflow_ref` for `playwright.yml`, including the `job_workflow_ref_semver_tag` form.
- Change `grafana/plugin-tools` to use the new actions. Re-enable its disabled `deploy-report-pages` job.
- `playwright-docker.yml` still uses a raw `actions/upload-artifact` step. Its artifact name does not match the pattern that `publish-report` reads. Docker-based runs therefore publish no reports.


tolzhabayev:
@mckn stale?

mckn:
hmm, not really - I should pick this up and get this one cleaned up and mereged.

sunker:
IIUC, this grants bucket writes per calling repo, so each plugin repo needs its own entry in deployment_tools. It's my understanding that the rest of this repo grants access to the shared workflow itself, not to the repos calling it. This I guess is also why the [onboarding requirements](https://enghub.grafana-ops.net/docs/default/component/grafana-plugins-platform/plugins-ci-github-actions/010-plugins-ci-github-actions/#requirements) have no deployment_tools step today and every plugin repo inherits GCS access for free? And it only covers pull requests, so a push to main or a nightly run has nothing to authenticate as, which means we lose those reports.

Could we not follow the same trust pattern here as for the rest of plugin-ci-workflows?

mckn:
> IIUC, this grants bucket writes per calling repo, so each plugin repo needs its own entry in deployment_tools. It's my understanding that the rest of this repo grants access to the shared workflow itself, not to the repos calling it. This I guess is also why the [onboarding requirements](https://enghub.grafana-ops.net/docs/default/component/grafana-plugins-platform/plugins-ci-github-actions/010-plugins-ci-github-actions/#requirements) have no deployment_tools step today and every plugin repo inherits GCS access for free? And it only covers pull requests, so a push to main or a nightly run has nothing to authenticate as, which means we lose those reports.
> 
> Could we not follow the same trust pattern here as for the rest of plugin-ci-workflows?

hmm, let me investigate if that would be possible. It might not be possible because I have changed to WIF authentication so I can limit access only to the folder in the bucket that the repository should have access to.

Basically each repo will get their own folder and they are only allowed to append files to that folder. Otherwise they would be able to alter other repositories reports which could be a pretty nasty way to inject malicious code to the reports.

I will get back on this - great feedback!

mckn:
> IIUC, this grants bucket writes per calling repo, so each plugin repo needs its own entry in deployment_tools. It's my understanding that the rest of this repo grants access to the shared workflow itself, not to the repos calling it. This I guess is also why the [onboarding requirements](https://enghub.grafana-ops.net/docs/default/component/grafana-plugins-platform/plugins-ci-github-actions/010-plugins-ci-github-actions/#requirements) have no deployment_tools step today and every plugin repo inherits GCS access for free? And it only covers pull requests, so a push to main or a nightly run has nothing to authenticate as, which means we lose those reports.
> 
> Could we not follow the same trust pattern here as for the rest of plugin-ci-workflows?

You are right, and thanks for catching it.

I changed grafana/plugin-ci-workflows#819 to follow that pattern. It impersonates `github-e2e-test-artifacts@` again, and it publishes on every event. I kept the dedicated service account rather than `github-plugin-ci-workflows@`, so the account still holds `objectUser` on one bucket only.

One trade-off for the record. Per-repository isolation now comes from the action instead of from IAM. But that should be fine according to our previous meeting 👍🏻

.github/workflows/ci.yml
```diff
@@ -366,6 +373,7 @@ concurrency:
 permissions:
   contents: read
   id-token: write
+  pull-requests: write
```
sunker, line 376: This is workflow-level, so I guess every job inherits it. `publish-reports` already sets its own job-level permissions, so could this go on just the `playwright:` job instead?

.github/workflows/playwright.yml
```diff
@@ -350,13 +358,33 @@ jobs:
           PLAYWRIGHT_CONFIG: ${{ inputs.playwright-config }}
         working-directory: ${{ inputs.plugin-directory }}
 
-      - name: Upload artifacts
-        uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
-        if: ${{ (inputs.upload-artifacts == true) && ((always() && steps.run-tests.outcome == 'success') || (failure() && steps.run-tests.outcome == 'failure')) }}
+      - name: Upload report artifacts
+        uses: grafana/plugin-ci-workflows/actions/plugins/playwright-reports/upload-report-artifacts@main
+        if: ${{ always() && inputs.upload-artifacts == true && (steps.run-tests.outcome == 'success' || steps.run-tests.outcome == 'failure') }}
```
sunker, line 375: Should this have the same fork check as `publish-reports` below?

.github/workflows/playwright.yml
```diff
@@ -350,13 +358,33 @@ jobs:
...
+          grafana-version: ${{ matrix.GRAFANA_IMAGE.VERSION }}
+
+  publish-reports:
+    name: Publish Playwright reports
+    needs: playwright-tests
+    # Runs on every event, including push and scheduled runs: the service account is trusted via
+    # job_workflow_ref, which says nothing about the event. Fork pull requests are excluded
+    # because GitHub withholds id-token: write from them. On a run with no pull request the action
+    # skips the comment steps and writes the results table to the job summary instead.
+    if: ${{ always() && !cancelled() && inputs.upload-artifacts == true && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.fork == false) }}
```
sunker, line 388: Could we use `isTrusted` from `needs.test-and-build.outputs.workflow-context` here? That's what the rest of ci.yml uses, and it also skips untrusted bots - as is, a Renovate PR would publish reports.

sunker: Should we also add `continue-on-error` on this job? Don't think reports should be able to fail a build or are they today?

actions/plugins/playwright-reports/publish-report/action.yml
```diff
@@ -0,0 +1,128 @@
...
+        use_wif_auth: 'false'
+        # Keep the credentials file for the upload step below, then drop it in "Revoke GCS
+        # credentials". Deleting it here would break the upload.
+        delete_credentials_file: 'false'
+
+    - name: Upload reports to GCS
+      uses: google-github-actions/upload-cloud-storage@6397bd7208e18d13ba2619ee21b9873edc94427a # v3.0.0
+      with:
+        path: all-reports
+        destination: ${{ steps.login-to-gcs.outputs.bucket }}/${{ github.repository }}/${{ steps.timestampid.outputs.timestamp }}/${{ steps.set-initiator.outputs.job_initiator }}
```
sunker, line 75: What's the date segment for? A re-run the next day writes to a fresh prefix, no? The PR number already makes the path unique I guess?

sunker (commented):
Overall looks really good, nice work! Added a few comments.

Some more high level feedback:
* We need some kind of docs on what this feature is and how to enable it. Perhaps put that in enghub next to other [plugin-ci-workflow docs](https://github.com/grafana/grafana-catalog-team/blob/main/docs/plugins-ci-github-actions/010-plugins-ci-github-actions.md)? 
* As said in the meeting, would be good to verify this works in one internal plugin and one external.

PR #881, opened 2026-07-30
fix: package every backend executable family

NWRichmond:
## What this changes

- Discover every backend executable family declared by a packaged `plugin.json`, resolving executable paths relative to that manifest.
- Build each tailored archive with the selected OS/architecture binary from every family and no binaries for other platforms.
- Reject mismatched backend platform matrices before creating incomplete tailored archives.
- Keep universal archive behavior and existing tailored archive naming unchanged.

## Why

The previous app-plugin path replaced the root executable family with a nested datasource executable. Tailored archives therefore filtered only the nested family and retained every platform binary from the root family.

This was observed in Odin's [development catalog publication job](https://github.com/grafana/hackathon-14-grafana-odin/actions/runs/30509388201/job/90767480397), where validation passed before the catalog API returned `InternalServer`. The oversized archives motivate this fix, but the catalog response alone does not establish that the failure was an OOM.

## Validation

- `go test -short -run '^TestPackageScript' -count=1 -v` (including macOS Bash 3.2)
- Direct existing single-backend validation: six tailored archives retain exactly one matching executable; universal retains all six
- `go vet ./...` in `tests/act`
- `shellcheck -e SC2086,SC2153 actions/internal/plugins/package/package.sh` (excluded findings predate this change)
- `git diff --check origin/main...HEAD`

Local environment limitations:

- `TestPackage` could not start because `act` is not installed.
- `make actionlint` and `make act-lint` could not start because `actionlint` and `golangci-lint` are not installed.

### DIFF
--- a/actions/internal/plugins/package/package.sh
+++ b/actions/internal/plugins/package/package.sh
@@ -77,60 +77,119 @@ if [ "$universal" = true ]; then
     exit 0
 fi
 
-# Identify apps with nested datasource
-ptype=$(jq -r .type plugin.json)
-exe=$(jq -r .executable plugin.json)
-backend_folder="."
-if [ "$ptype" == "app" ] && [ -d "datasource" ]; then
-    echo "Found nested datasource"
-    cd datasource
-    nested_exe=$(jq -r .executable plugin.json)
-    if [ "$nested_exe" != "null" ]; then
-        backend_folder="datasource"
-        exe="$backend_folder/$nested_exe"
+# Discover every backend executable family declared by a packaged plugin.json.
+# Executable paths are relative to the plugin.json that declares them.
+backend_executables=()
+while IFS= read -r -d '' plugin_json; do
+    executable=$(jq -r '.executable // empty' "$plugin_json")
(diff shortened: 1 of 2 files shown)


NWRichmond:
CI note: the Act jobs did not execute this patch. The [checkout step](https://github.com/grafana/plugin-ci-workflows/actions/runs/30511584700/job/90772654385) fetched the upstream PR checkout, then ran `git checkout fix/multi-backend-plugin-packaging`; that branch exists only in the fork, so Git reported `pathspec ... did not match` and all test steps were skipped.

The focused direct package-script regressions, including multi-backend tailoring, mismatched-matrix rejection, universal contents, Windows suffixes, and a nested executable path, pass locally under macOS Bash 3.2. The existing single-backend mockdata also produces the unchanged six tailored archive names/contents plus one universal archive.

tolzhabayev:
duplicate / same area of https://github.com/grafana/plugin-ci-workflows/pull/882

NWRichmond:
> duplicate / same area of #882

Thanks! Good timing, #882 will help :)

PR #883, opened 2026-07-30
test(act): cover packaging of an app with a nested datasource backend

tolzhabayev:
# What

Adds a `simple-app-nested-backend` test fixture (an app with its own backend AND a nested datasource backend, the grafana-odin-app shape) and a third `TestPackage` case asserting that each os/arch zip contains exactly one binary per backend. The existing strict zip-content checks would have caught the packaging bug fixed in #882 immediately, but no fixture exercised either nested-datasource path.

**This PR depends on #882 and its act tests are expected to fail until #882 merges.** That's deliberate: the red run demonstrates the bug. Verified locally both ways:

- on this branch (without the fix), the new case fails with the leaked binaries:
  ```
  unexpected file "/grafana-simpleappnested-app/gpx_simpleappnested_app_darwin_arm64" found
  unexpected file "/grafana-simpleappnested-app/gpx_simpleappnested_app_linux_amd64" found
  unexpected file "/grafana-simpleappnested-app/gpx_simpleappnested_app_linux_arm" found
  ...
  ```
- with the #882 fix cherry-picked on top, it passes

After #882 merges, rebase this on main and CI goes green.

## Changes

- `tests/simple-app-nested-backend/`: minimal fixture. The frontend is static files copied by `build.js` on purpose (no webpack toolchain, packaging is what's under test). The two Go backends are dependency-free `main` packages built by a custom `mage buildAll` into the same dist layout the SDK produces, so `make mockdata` keeps working for it.
- `TestPackage`: the table now carries per-case `expExecutables` (path prefixes of each backend binary) and `expExtraDistFiles` instead of hardcoding `gpx_simple_backend`. Existing cases are unchanged in behavior.
- Mockdata for the fixture. The `dist-artifacts-unsigned` zips were generated with the fixed package.sh from #882, since that's what `make mockdata` reproduces once it merges.


tests/act/main_package_test.go
```diff
@@ -21,6 +21,13 @@ func TestPackage(t *testing.T) {
 		expPluginVersion string
 		expBackend       bool
 		expPluginType    string
+		// expExecutables lists each backend executable's path prefix inside
+		// the plugin folder, without the os/arch suffix. Per-platform zips
+		// must contain exactly one binary per entry.
```
xnyo:
```suggestion
		// contain exactly one binary per entry 
		// (unless it's a nested plugins where both the parent and child contain nested backend executables)
```

xnyo (commented):
Just one small comment the rest LGTM!

xnyo (approved):
LGTM!

PR #1016, opened 2026-09-09
fix(setup): cache the pnpm store from the plugin directory

tolzhabayev:
The act tests have been failing on `TestSmoke/simple-frontend-pnpm` and `TestToolingVersions/simple-frontend-pnpm` since around 7 September as pnpm 12 was released.

`setup-node` finds the pnpm cache by running `pnpm store path` from the repository root. Our test plugins live in `tests/simple-*`, so there is no package.json at the root to pin a pnpm version, and corepack answers with whatever is newest on the registry. Up to pnpm 11 that returned the global store, which exists after an install. pnpm 12 returns a path under the current directory instead, so the probe reports a store at the repo root while the install creates one in the plugin folder. The cache-save post step then fails path validation and fails the job.

So this caches the pnpm store directly and resolves it in the plugin directory, where the packageManager field pins the version. The result no longer depends on which pnpm happens to answer. The pnpm cache also starts working, it has been silently missing on every run.

Verified locally with `go test -run TestSmoke\|TestToolingVersions`, both green, and green again with the stale corepack pnpm shim put back in the act toolcache volume so it matches what the runners have.

### DIFF
--- a/actions/internal/plugins/setup/action.yml
+++ b/actions/internal/plugins/setup/action.yml
@@ -99,8 +99,27 @@ runs:
       with:
         node-version: "${{ inputs.node-version }}"
         node-version-file: "${{ inputs.node-version-file }}"
-        cache: ${{ inputs.act-cache-warmup != 'true' && steps.package-manager.outputs.name || '' }}
+        # pnpm is cached separately below. setup-node looks the store up by running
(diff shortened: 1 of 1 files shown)


andresmgot (approved):

xnyo (approved):
LGTM!

